PR #23, opened 2021-03-18
Test rework

alfredclwong:
These changes are just suggestions - I'm putting them into a PR for review/discussion.

@BrianGun I remember you saying that you don't use our custom Precompile/Sysimage code anymore - if this is the case, should we clean these files? Similarly, I don't see usages for Miscellaneous.jl (aka Junk) or Spectra.jl. Let me know if they're still here for a reason and I'll revert the specific commits.

Specifying a separate test/Project.toml helps us to lighten the core dependency list.

The big change: I think Benchmarks, Diagnostics and TestData naturally belong in test/. Some docs/examples use TestData, but otherwise they're largely separate from the core functionality and they take up a good chunk of compile time. We can just include($path_to_TestData) for the few exceptions, as I've done here. What do you guys think?


BrianGun:
moving the files is okay. I wouldn't remove the precompile file. It executes some visualization code during precompilation that makes time to first render faster. The build task in VSCode doesn't do this. 

The reason I stopped using it was because Julia Computing support hypothesized that some corrupt type information generated during the precompilation caused by the precompile file was causing an obscure bug in our code. But we haven't had any trouble like that for a while so it was probably a transient thing.

BrianGun (approved):

PR #41, opened 2021-03-22
Glasscat changes

### LINKED ISSUE #37: GlassCat tests
Working on incorporating these into the main test code as a test set.

alfredclwong:
Main purpose of this PR is to incorporate GlassCat tests (resolves #37). Also includes some refactoring.

~~WIP: regression tests failing (see #37)~~

WIP: windows tests failing due to lack of OpenGL support. Thought I'd fixed this with the Mesa install - looking into it now.


friggog:
Would be good to keep all .jl files with uppercase names for consistency

alfredclwong:
The convention I'm going for is that files which define a type/module are uppercase and everything else is lower.

I'm doing this because the main module file OpticSim.jl is required to be uppercase, but standard scripts such as test/runtest.jl and deps/build.jl are required to be lowercase. imo this is the most consistent approach, but it's debatable.

What do you think? Either way I'll have to rename a few other files to match.

friggog:
Ah ok - yeah that sounds sensible, as long as we are consistent throughout. 
I'm not sure if there is any general Julia guidance on this which we should follow?

alfredclwong:
From what I've seen, there's no established convention. [Revise](https://github.com/timholy/Revise.jl/tree/master/src) uses all lower, [Pluto](https://github.com/fonsp/Pluto.jl/tree/main/src) all upper, and [Pkg](https://github.com/JuliaLang/Pkg.jl/tree/master/src) uses a mix.

Will get round to changing the other filenames soon.

BrianGun (approved):

PR #47, opened 2021-03-24
Galran/new emitter

galran:
Added the new emitters implementation and some Pluto Notebooks utilities and examples.


BrianGun (approved):

alfredclwong (changes requested):
This is a big PR so I have a lot of review comments. If the tone seems terse it's for the sake of brevity: nothing personal! Here goes...

https://github.com/microsoft/OpticSim.jl/pull/47/commits/16e5de40904962daa987f628f00355bb93fb6925

**OpticSim.jl**
What do you mean by "a better approch to how and where visualization code should reside"? Would you prefer for visualization code to be local to the corresponding logical code? I'm not sure this is a good idea.

**Emitter4.jl**
There are a lot of generic utility definitions, e.g. `MissingImplementationException`, `@timeCap`, etc. Could you move these out of the submodule?

Along the same lines, shouldn't Emitters.Geometry and Emitters.Visualization be merged with the existing OpticSim submodules?

There is a switch from relative imports (`import ..OpticSim`) in `OpticSim.Emitters` to `import OpticSim` in the triple nested submodules (e.g. `OpticSim.Emitters.Geometry`). This seems inconsistent - is `import ...OpticSim` possible?

`unitX`/`unitY`/`unitZ` are defined in the local scope, but `OpticSim.origin` is defined in `OpticSim`. Why?

Is it necessary to force a specific type (via `::Type{T} = Float64`) in `rotXmat` etc.? This seems to defeat the purpose of parametric typing.

On the other hand, `Emitters.generate(s::Uniform{T}) where {T}` (L304) is missing some type checking: `T<:Real` seems minimal. And `Base.length(d::Constant{T}) where {T} = 1` (L430) doesn't use `T` - in this case, it's possible to omit `T` and get the same effect.

Aside from additional constructors, what's the difference between `Origins.Point` (L597) and `Geometry.Vec3D`?

Could you write docstrings for the new definitions? We definitely want information on Emitters in the auto-generated documentation.

https://github.com/microsoft/OpticSim.jl/pull/47/commits/63612b9ec0574d8f959aa66ede6a2655d30c9fd8

**Project.toml**
[compat] FileIO 1.6 -> 1.4. Was 1.6 causing issues?

**Misc/Misc.jl**
"Misc" is ambiguous. Could we instead call the folder something like "Notebook", containing the OpticSim.Notebook module?

**NotebooksUtils.jl**
Abbreviation (NB) is discouraged in Julia - better to use `const NB = OpticSim.Notebook` to define an alias. 

https://github.com/microsoft/OpticSim.jl/pull/47/commits/7175e188ba2629d9d4dad80bc0c9f1a3077aa439 (merge commit)

**Project.toml**
[compat] HTTP 0.9 -> 0.8. Why?

Tests! OpticSim is currently ~6000 LOC (lines of code), and Emitter4.jl alone is ~1000 untested LOC. This will drop our coverage way down to ~40%.

The #region comment blocks are pretty nifty. However, I think these represent natural splits into separate source files. Splitting achieves the same goal (encapsulation) in a more explicit and maintainable manner. Since Emitters is now a (large) submodule, these files would go in a folder Optical/Emitters/.

galran:
Lots of good comments 😊
I’m still figuring out some of them myself. I think Visualization needs to be as local as possible and I also think we should create a mechanism to localize the tests too using some consistent interfaces.
I think the CSG and Emitters Geometry should be merged asap.
There is a switch from relative imports (import ..OpticSim) in OpticSim.Emitters to import OpticSim in the triple nested submodules (e.g. OpticSim.Emitters.Geometry). This seems inconsistent - is import ...OpticSim possible? CORRECT!

…and so on…Just need more time lol

ran

alfredclwong:
I'm beginning work on these changes on the [test/new-emitter](https://github.com/microsoft/OpticSim.jl/tree/test/new-emitter) branch.

PR #48, opened 2021-03-24
Renaming

alfredclwong:
As discussed with @friggog, we were beginning to have file naming inconsistencies (my fault). This stemmed from the fact that standard Julia scripts such as `deps/build.jl` and `test/runtests.jl` are required to be lowercase, whereas our main module file `src/OpticSim.jl` is required to be CamelCase (uppercase).

This is my suggestion for a consistent naming convention:

- Files which define a module are named after the module (e.g. OpticSim.jl, Geometry.jl).
- Submodules are placed within their own folder with the same name (e.g. src/Geometry/Geometry.jl). This allows for submodules to be split into several source files. Ideally, the parent file can then just be a list of imports and includes (see [OpticSim.jl](https://github.com/microsoft/OpticSim.jl/blob/main/src/OpticSim.jl)).
- Test sets (e.g. test/testsets/JuliaLang.jl) are Camelcase.
- "Scripts" (files which only define functions) are given lowercase names. For example, we have various utilities.jl and constants.jl files that fit this description.
- All other files are CamelCase. Essentially, this means that all files are CamelCase with the exception of scripts.

I've tried to update existing names across the project to conform to these rules. Open to suggestions.

Also, I think we should bump the patch number to v0.3.1.


BrianGun:
Matrix.jl defines functions that aren't used anywhere.  @alfredclwong could update your commit to remove this file (after verifying I'm correct of course).  Then I'll approve. Everything else looks fine.

BrianGun:
@alfredclwong I'm not seeing my comment so repeating it here. Matrix.jl defines functions that aren't used anywhere. We should delete it from the project. Could you update your commit to do this? Everything else looks fine. I'll approve as soon as Matrix.jl is deleted.

BrianGun (commented):
@alfredclwong  it looks like matrix.jl is still being included in OpticSim.jl which will cause an error, since it is now deleted. If this is correct then OpticSim.jl needs to be edited.

friggog:
> Files which define types are camelcase. This is more arbitrary as it's sort of in between a module file and a script file.
Seems a little odd to me, are there many examples of this?
What if a file defines classes and functions but isn't a module?
I would be tempted to say upper case for everything except 'scripts' which should be lower

> Submodules are placed within their own folder with the same name
This is a good convention I think

alfredclwong:
> > Files which define types are camelcase. This is more arbitrary as it's sort of in between a module file and a script file.
> 
> Seems a little odd to me, are there many examples of this?
> What if a file defines classes and functions but isn't a module?
> I would be tempted to say upper case for everything except 'scripts' which should be lower

My wording wasn't very clear but I think we mean the same thing. I basically meant "all the other files which aren't modules or scripts" should be uppercase (CamelCase).

BrianGun (approved):

PR #66, opened 2021-03-29
Fix UndefVarError: sources not defined

AndiMD:
The `sources` variable does not exist. Apparently it was replaced by `raygenerator` (?)

```
ERROR: UndefVarError: sources not defined
Stacktrace:
 [1] trace(system::CSGOpticalSystem{Float64, Float32, Ellipse{Float64}, LensAssembly}, raygenerator::UniformOpticalSource{Float64, RandomGaussianOriginPoints{Float64}, RandomSource{Float64, RandomGaussianOriginPoints{Float64}}}; printprog::Bool, test::Bool, outpath::Nothing)
   @ OpticSim ~/.julia/packages/OpticSim/4qItW/src/Optical/OpticalSystem.jl:353
```

### DIFF
--- a/src/Optical/OpticalSystem.jl
+++ b/src/Optical/OpticalSystem.jl
@@ -350,9 +350,9 @@ function trace(system::CSGOpticalSystem{T}, raygenerator::OpticalRayGenerator{T}
         if k % update_timesteps == 0
             total_traced += update_timesteps
             dif = round(time() - start_time, digits = 1)
-            left = round((time() - start_time) * (length(sources) / total_traced - 1), digits = 1)
+            left = round((time() - start_time) * (length(raygenerator) / total_traced - 1), digits = 1)
             if printprog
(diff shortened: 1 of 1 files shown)


BrianGun (changes requested):
@AndiMD thank you for finding this bug. 

The first function in your PR, trace, does have sources undefined and errors on execution. 

The other two functions, traceMT, and tracehitsMT, have sources defined and do not error on execution. At least they don't error on my machine. Just tested all three functions.

Here's a quick test you can run to verify that traceMT and tracehitsMT work on your machine:

rays = RayListSource([OpticalRay([0.0,0.0,0.0],[0.0,0.0,1.0],1.0,.78) for _ in 1:
1000])
conv = Examples.doubleconvex()

#this should error
trace(conv,rays) 

#this should not error
traceMT(conv,rays)

#this should not error
tracehitsMT(conv,rays)

Could you change your PR to remove the changes to the latter two functions?

AndiMD:
thank your response, I do not want to sign a contract.

Issue #69, opened 2021-03-31
Use relative imports in AGFGlassCat.jl


friggog:
Currently the absolute path is written to AGFGlassCat.jl, it would be better to just use relative paths e.g. "NIKON.agf" for the includes.

It would also be great to have some clarity around the install/build process to help with setup on AML. I have jobs (almost) working when running from a clone of this repo, but ideally we'd (also) support downstream packages with OpticSim as a dependency, in which case we'll need to install OpticSim through Pkg on the VM. Will the build just happen automatically in this case and generate the glass stuff as needed?

alfredclwong:
Been busy with hardware stuff so I haven't had much time for OpticSim lately! Although I believe I did manage to write a short description of the build process in the docs. It's also standard Julia - builds, including downstream packages, will be triggered by Pkg.build or Pkg.add.

I can change to relative paths in the build script. This would mean passing "data" instead of JL_DIR to generate_jls() in build.jl L45. You can see why I went with absolute paths, as we also run mkdir() from deps/ in order to ensure the data directory exists. Is this causing an issue in cloud runs?

friggog:
Seems even simpler: https://github.com/microsoft/OpticSim.jl/pull/71/commits/3ff429438248a8c9f729b67a91efa39617f0fee8

The issue was only when running from a clone of this repo as the source files are uploaded as a snapshot meaning the absolute paths broke stuff.

Building seems to work fine on the cloud runs when OpticSim is a dependency, I sometimes get errors in the logs (and locally) for build/precompilation but it doesn't seem to affect things and it's not clear what package is the issue?

alfredclwong:
Ah yes I forgot that AGFGlassCat.jl was in data/ already, thanks.

That's what I'm a bit confused about - clones from the repo shouldn't contain AGFGlassCat.jl. Is there a central controller that clones the repo, builds the package, and then distributes the generated code to workers?

I see those ambiguous errors too in Julia 1.6. Not sure how to find out which package is at fault.

friggog:
I cloned it to my local machine then ran `] build` to generate the files so I can open the REPL with `--project` inside the OpticSim.jl directory and use the package from there. With AML a snapshot of your "source directory", in this case my cloned OpticSim.jl directory, is uploaded and the script executed within that - hence the problem.

If OpticSim is a dependency within a separate project where that project's source directory is uploaded then there isn't a problem because OpticSim gets installed through Pkg on the VM and built there.

[...]

Issue #40, opened 2021-03-22
Make it easier to run on the cloud


friggog:
It would be great if it was easier to run large simulations on the cloud.
It would be pretty easy to set up a custom docker image with Julia installed and ideally a sysimage there too with the OpticSim dependencies in - this would be hugely valuable in itself for cloud execution.

To go even further we could then have a script within the default package which submits runs e.g. to Azure ML using this docker image and whatever command the user wants to execute.
AzureML has a nice thing where it uploads a snapshot of the source directory, so it is possible to modify the source code and run jobs with those modifications.

e.g.
```julia
OpticSim.submit_job_to_AML(,aml_workspace, relpath_to_entry_script, args_for_script)
```

I doubt there is a julia wrapper for AML API, but we could probably call the python API somehow...

Certainly not sure on the details here, but enabling cloud runs seems like something that should be reasonably high priority

alfredclwong:
This ties in nicely with our push towards Pluto notebooks as we won't want to do heavy (or even light) simulations on a web server. I'll look into it when I can.

BrianGun:
@friggog were you going to take a crack at this? Sounds like you have the most knowledge about AzureML.

friggog:
Yeah I’m happy to take a look, at least to get the basics running and then maybe you can add a more comprehensive framework on top of that

friggog:
Resolved by #71

PR #71, opened 2021-04-01
AML cloud support

friggog:
Adds support for running simulations on AML, addressing #40 

TODO:
- [x] Needs #69 for this to work when running from a clone of the repo
- [x] Hyperdrive not currently supported
- [x] Figure out outputs - what is the typical output we want/how do we transfer back from AML?
- ~~_No sysimage_ - docker has packages built/precompiled when using OpticSim as a dependency which is probably sufficient~~
- [x] Add docs

We could call julia directly using ScriptRunConfig "command" rather than having the entry script stuff, I think we'll always need to resolve the environment first though and it becomes a pain to configure thread count.


alfredclwong:
re outputs: I was thinking of writing some kind of "Ray Database" type, similar to a .ZRD file in Zemax, maybe next week. Any other thoughts?

BrianGun (commented):
@friggog do you have a simple example you could add to the Examples directory? Maybe a call to traceMT which returns an image? Since any example would need an Azure account to bill against that part could be left for the user to fill in, perhaps with inline comments that explain what the user needs to do to make it work. And maybe a paragraph of documentation in the docs? You could add a new section, Cloud Computation, or something like that. Otherwise people are going to have a hard time figuring out how to use this.

BrianGun:
@friggog were you going to add an example and some documentation? It will be confusing to new users otherwise.

friggog:
Yeah there are a few things to do still (see list on top comment here) hopefully will get around to it this week but will see how busy I am

BrianGun:
Charlie,

You should have permission to approve pull requests. Could you verify for me?

-brian

friggog:
@alfredclwong can I preview built docs from this branch somehow?

alfredclwong:
@friggog deploydocs has an argument that'll push PR previews to the gh-pages branch. i've pushed a commit that'll try out that feature.

alfredclwong (approved):
Thanks for taking the time to do this. I don't think I have the permissions yet to try it out, but I assume you've tested it yourself.

docs/src/cloud.md
```diff
@@ -0,0 +1,51 @@
...
+
+- Subscription ID, of the form `XXXXXXXX-XXX-XXX-XXXX-XXXXXXXXXXXX`
+- Resource group name
+- Workspace name
+- Compute cluster name
+
+This information can be cached either to a specific file, or globally:
+
+```@docs
+Optics.Cloud.cache_run_config
```
alfredclwong, line 20: Need to change Optics -> OpticSim

alfredclwong (changes requested):
3 lines need changing - I'll push the changes in a sec

docs/src/cloud.md
friggog: Yeah, publishing works great - definitely a useful feature https://microsoft.github.io/OpticSim.jl/previews/PR71/cloud/

friggog:
> Thanks for taking the time to do this. I don't think I have the permissions yet to try it out, but I assume you've tested it yourself.

yeah it seems to be working for a simple test script. The pycall setup is the most dodgy bit so maybe need some refinement there, seems to be very dependent on local environment (ie conda/python install)

I might be able to give you permissions though I just removed my ownership so we might need Ran or Brian to do that now

alfredclwong (approved):

PR #83, opened 2021-04-07
Fix new emitters

alfredclwong:
[WIP]

I didn't get a chance to review PR #79 as I was off on holiday. Having a look at it now, several things slipped through the code review. These are the most obvious fixes - when I have more time, I may have a deeper look. There are also some issues from PR #70.

- [x] We now have Emitter.jl, Emitter2.jl and Emitters.jl in the main branch. I don't agree with the decision to make dealing with this a low-priority matter, so I'll get this fixed here (delegated to Issue #104).
- [x] Some type signatures need fixing. For example, in https://github.com/microsoft/OpticSim.jl/blob/88a22f55d07ff5f274243602d2e913c7d5701d34/src/Optical/Emitters.jl#L24-L27 the parametric type `T` is unused, and in https://github.com/microsoft/OpticSim.jl/blob/88a22f55d07ff5f274243602d2e913c7d5701d34/src/Optical/Emitters.jl#L222 and a few other cases, it's entirely redundant.
- [x] Several relative import statements don't seem correct. For example, https://github.com/microsoft/OpticSim.jl/blob/88a22f55d07ff5f274243602d2e913c7d5701d34/src/Optical/Emitters.jl#L36-L40 doesn't make sense to me, as OpticSim and Emitters should be on different levels of nesting.
- [ ] Stylistically, Vec3/Vec4 may need some discussion. Since we're working with computer graphics, I'm not wholly opposed to explicit Vec3/Vec4 types, but some of the implementations, e.g. https://github.com/microsoft/OpticSim.jl/blob/88a22f55d07ff5f274243602d2e913c7d5701d34/src/Geometry/Transform.jl#L44 are redundant due to LinearAlgebra (here, `ones(Vec3{T})` would suffice). I think there are also some instances where the type signatures could make better use of multiple dispatch.
- [x] Spelling/grammar mistakes.
- [x] In https://microsoft.github.io/OpticSim.jl/dev/csg/, csg_ex.png is broken due to `rotationd` not being in scope. I'll change the doc build Github Action to fail on things like this now (I wasn't doing this because of the zoom lens example, but I'll comment that out).
- [x] Docstring errors (delegated to Issue #103).
- [x] What is this struct for? @galran https://github.com/microsoft/OpticSim.jl/blob/88a22f55d07ff5f274243602d2e913c7d5701d34/src/Optical/Emitters.jl#L11-L13
- [x] Emitters testset (delegated to Issue #104).

Further issues:

- [ ] Redefinition of const parametric type T. @galran what's the intended purpose here? https://github.com/microsoft/OpticSim.jl/blob/103c70fb7391502a33cccc5c98531a832dcb93b0/src/Optical/Emitters.jl#L120
- [x] export Geometry called from Emitters module. This is exporting an imported module. https://github.com/microsoft/OpticSim.jl/blob/103c70fb7391502a33cccc5c98531a832dcb93b0/src/Optical/Emitters.jl#L1082
- [x] export generate called from Directions and Origins submodules. https://github.com/microsoft/OpticSim.jl/blob/103c70fb7391502a33cccc5c98531a832dcb93b0/src/Optical/Emitters.jl#L188 https://github.com/microsoft/OpticSim.jl/blob/103c70fb7391502a33cccc5c98531a832dcb93b0/src/Optical/Emitters.jl#L408
- [ ] Looking at these lines again, it seems that they should actually reside in Transform.jl. @galran is OpticSim.origin extended explicitly here because some previous code uses it? https://github.com/microsoft/OpticSim.jl/blob/88a22f55d07ff5f274243602d2e913c7d5701d34/src/Optical/Emitters.jl#L24-L27


src/Optical/Emitters/Directions.jl
```diff
@@ -0,0 +1,206 @@
...
+    end
+    
+    function Constant(::Type{T} = Float64) where {T<:Real}
+        direction = unitZ3(T)
+        return new{T}(direction)
+    end
+end
+export Constant
+
+Base.length(d::Constant{T}) where {T} = 1
```
BrianGun, line 47: this use of length is not consistent with general Julia practice, which uses length to indicate the number of elements in a collection. A better name would be norm which is the pnorm of a vector:

Base.norm(d::Constant{T},pnorm=2) = norm(d.direction,pnorm)

src/Optical/Emitters/Directions.jl
```diff
@@ -0,0 +1,206 @@
...
+    dir = cos(θv) * (cos(θu) * direction + sin(θu) * uvec) + sin(θv) * vvec
+    return dir
+end
+
+
+#---------------------------------------
+# Cone Ray Distribution
+#---------------------------------------
+"""
+Encapsulates a numsamples rays sampled uniformlly from a cone with max angle θmax.
```
BrianGun, line 105: uniformly

src/Optical/Emitters/Directions.jl
```diff
@@ -0,0 +1,206 @@
...
+    
+    function Constant(::Type{T} = Float64) where {T<:Real}
+        direction = unitZ3(T)
+        return new{T}(direction)
+    end
+end
+export Constant
+
+Base.length(d::Constant{T}) where {T} = 1
+function Emitters.generate(d::Constant{T}, n::Int) where {T<:Real}
```
BrianGun, line 48: It seems strange to define an Emitters function inside the Directions module. This is going to make it hard for users to find all relevant function definitions, since they will be scattered in unlikely places. 

Why is this necessary?

src/Optical/Emitters/Directions.jl
```diff
@@ -0,0 +1,206 @@
+module Directions
+
+using ....OpticSim, ...Geometry
```
BrianGun, line 3: I understand the desire to encapsulate definitions to avoid polluting the namespace but this may be taking it too far. Is it really necessary for these modules to be nested so deeply? This is not considered good Julia programming style.

src/Optical/Emitters/Directions.jl
alfredclwong: @BrianGun feel free to push small changes like this directly to the branch. Comments are also fine - it'll just take me a bit longer to get round to it.

src/Optical/Emitters/Directions.jl
alfredclwong: From my understanding, this is a good solution if we are going with the submodule approach.

Essentially, Emitters has "generate()" in its interface, and this can be called within various contexts, such as for a hexapolar origin or composite source.

src/Optical/Emitters/Emitters.jl
```diff
@@ -0,0 +1,41 @@
...
+module Emitters
+
+using ...OpticSim, ..Geometry
+using LinearAlgebra
+
+struct MissingImplementationException <: Exception
+    msg::String
+end
+
+# defining name placeholders to override in nested modules
```
BrianGun, line 15: Now I understand why these functions are defined here because otherwise when they are overriden in other modules you will get a compiler error saying the function is not defined.

This seems like a **BAD** idea. Not standard Julia style, hard to understand, scatters the function definitions for a module across other unrelated modules. If there are multiple definitions for the same function in multiple modules (let's hope this never happens) the resulting function definition will be the last one the compiler executes. This is difficult for a developer to predict since it requires tracking in what order the compiler compiles the modules.

These modules should be re-architected to eliminate the need to do this.

BrianGun (commented):
The deep nesting of modules still seems unnecessarily complicated to me but in the interests of making progress let's try this out and see how it works.

src/Optical/Emitters/Directions.jl
BrianGun: Sorry, misread the code. This seems okay given that we want modules nested so deeply.

src/Optical/Emitters/Directions.jl
alfredclwong: Just had a proper look at the code and it seems like Base.length is being used appropriately here.

alfredclwong:
@galran I've now set the documentation build to fail on strict requirements, so we'll need to fix all the missing docstrings. This is the current list:

```
┌ Error: 15 docstrings not included in the manual:
│ 
│     OpticSim.Geometry.local2world :: Union{Tuple{StaticArrays.SMatrix{4, 4, T, 16}}, Tuple{T}} where T<:Real
│     OpticSim.Geometry.decomposeRTS :: Union{Tuple{StaticArrays.SMatrix{4, 4, T, 16}}, Tuple{T}} where T<:Real
│     OpticSim.GlassCat.CARGILLE.OG0607
│     OpticSim.Geometry.world2local :: Union{Tuple{StaticArrays.SMatrix{4, 4, T, 16}}, Tuple{T}} where T<:Real
│     OpticSim.NotebooksUtils.SetBackend :: Tuple{OpticSim.NotebooksUtils.Defs, String}
│     OpticSim.Emitters.Spectrum.spectrumpower :: Union{Tuple{T}, Tuple{OpticSim.Emitters.Spectrum.Measured{T}, T}} where T<:Real
│     OpticSim.GlassCat.CARGILLE.OG0608
│     OpticSim.NotebooksUtils.run :: Tuple{}
│     OpticSim.NotebooksUtils.run :: Tuple{Any}
│     OpticSim.NotebooksUtils.InitNotebook :: Tuple{}
│     OpticSim.Geometry.scale :: Tuple{T} where T<:Real
│     OpticSim.Geometry.scale :: Union{Tuple{T}, Tuple{T, T, T}} where T<:Real
│     OpticSim.Geometry.scale :: Union{Tuple{OpticSim.Geometry.Vec3{T}}, Tuple{T}} where T<:Real
│     OpticSim.GlassCat.CARGILLE.OG081160
│     OpticSim.NotebooksUtils.run_sample :: Union{Tuple{String}, Tuple{String, Bool}}
│ 
│ 
│ These are docstrings in the checked modules (configured with the modules keyword)
│ that are not included in @docs or @autodocs blocks.
└ @ Documenter.DocChecks ~/.julia/packages/Documenter/bFHi4/src/DocChecks.jl:69
```

Could you please add references to these into the appropriate .md files in docs? Thanks.

@BrianGun, the CARGILLE docstrings are a bit awkward. I think our two best options are either (1) to include a documentation page for CARGILLE, or (2) to remove CARGILLE entirely - currently it's only used once, in a testset.

friggog:
@alfredclwong if you are no longer providing docs for all glasscat catalogs (which seems like a sensible choice) then it’s probably best to just remove the docstrings from cargille, there shouldn’t be any need to remove the code

alfredclwong:
> @alfredclwong if you are no longer providing docs for all glasscat catalogs (which seems like a sensible choice) then it’s probably best to just remove the docstrings from cargille, there shouldn’t be any need to remove the code
> 
> Also updating the [complete reference page](https://microsoft.github.io/OpticSim.jl/stable/ref/) in the docs with new sub modules (using auto docs) will remove these warnings if you don’t want dedicated docs pages for all these

Removing the docstrings seems like the way to go. Or at least making them into multi-line comments. @BrianGun, do you have any objection to this? There is a tiny bit of convenience in seeing the glass info when you hover over a glass in the source code.

autodocs also sounds good. @galran, did you want to put those docstrings in new_emitters.md at all? If not, I'll just add them to ref.md.

BrianGun:
> > @alfredclwong if you are no longer providing docs for all glasscat catalogs (which seems like a sensible choice) then it’s probably best to just remove the docstrings from cargille, there shouldn’t be any need to remove the code
> > Also updating the [complete reference page](https://microsoft.github.io/OpticSim.jl/stable/ref/) in the docs with new sub modules (using auto docs) will remove these warnings if you don’t want dedicated docs pages for all these
> 
> Removing the docstrings seems like the way to go. Or at least making them into multi-line comments. @BrianGun, do you have any objection to this? There is a tiny bit of convenience in seeing the glass info when you hover over a glass in the source code.
> 
> autodocs also sounds good. @galran, did you want to put those docstrings in new_emitters.md at all? If not, I'll just add them to ref.md.

I don't have any objection to removing the docstrings.

BrianGun (dismissed):

BrianGun (dismissed):

BrianGun:
@alfredclwong documentation check is failing.

alfredclwong:
@BrianGun please don't merge main into this branch using Github. There are some things in main that I don't want here, such as the original Emitters.jl file. I just realised that one of my merges from earlier reintroduced it.

alfredclwong:
In 1916e2e, I'm not sure if Int or Integer is preferable. I chose Integer because it's more generic, but it might decrease performance because of this. @galran @BrianGun @friggog any thoughts?

friggog:
@alfredclwong it might be a problem, `Integer` is an abstract type (supertype of all ints) so may not have known size to the compiler, possibly resulting in allocations. `Int` is an alias for `Int32` or `Int64` depending on the system so is always of known size to the compiler. It might be ok, but make sure you check carefully whether any runtime allocations have been introduced

alfredclwong:
> @alfredclwong it might be a problem, `Integer` is an abstract type (supertype of all ints) so may not have known size to the compiler, possibly resulting in allocations. `Int` is an alias for `Int32` or `Int64` depending on the system, so is always of known size to the compiler. It might be ok, but make sure you check carefully whether any runtime allocations have been introduced

That's what I'm worried about. I feel like having abstract types in function signatures is OK, but there might be performance issues if we allow struct fields to be too generic. What's the best way to test for allocations - adding representative workloads to Benchmarks.jl?

alfredclwong:
@galran do you have any recommendations for a representative workload for the Emitters code?

BrianGun:
There are several things to consider. First what is the intent of the type? Here are the subtypes of Integer:

Bool
 Signed <- ( BigInt, Int128, Int16, Int32, Int64, Int8) where <- means the elements on the right are subtypes
 Unsigned <- (UInt128, UInt16, UInt32, UInt64, UInt8)

Will this code even work properly when passed all of these subtypes? Does it make sense to have this degree of generality? If the answer is no then it shouldn't be Integer.

Then there is the performance consideration. In general giving struct fields abstract types will lead to poor performance for the reasons @friggog mentioned. In this particular case Int64 is probably a better idea than Integer. No performance hit, no allocation, and plenty big enough to represent enough rays to keep a computer busy for a long time (billions of seconds, at least).

alfredclwong:
I'll revert the commit in a bit

galran:
regarding the emitters, the only thing i can think of is running a couple of the examples (from the documentation or the notebook) - or at least just instantiating the non-visual part in order to not drag mesa into the mix.

BrianGun (approved):

PR #95, opened 2021-04-08
Update [compat] entries for new dependencies

alfredclwong:
Going forward, don't forget to add [compat] entries for new dependencies, as otherwise our package might be ineligible for the General Registry.


friggog (approved):

friggog:
My bad!

PR #96, opened 2021-04-08
Fix sources

alfredclwong:
1. NIKON link changed
2. HOYA website ([hoyaoptics.com](http://hoyaoptics.com/)) is down. New link found - @BrianGun could you verify that this link is OK? Added support in build script for direct AGF downloads (previously only supported ZIP archives).
3. Sumita link changed. @BrianGun there's a download option for discontinued glasses - do we want this?

Also extended ZIP file support to automatically pick the first AGF file in the downloaded ZIP archive.


BrianGun (approved):

BrianGun:
Don't think we want to include discontinued Sumita glasses, or other manufacturers, in the default download. Let users get them in a separate step if they need them.

PR #99, opened 2021-04-08
Speed up tests

### LINKED ISSUE #54: Speed up slow tests
Test timings are similar across Julia versions (1.5-1.7) and OS's (windows/ubuntu). In 1.7, @time shows that SurfaceDefs and Lenses spend ~20% of their time on compilation. Pending JuliaLang testset timings.

Julia 1.7 runs pre-compilations in the build step. We can see that windows/ubuntu takes 9/12 mins, respectively, on this. I think this block of time is irreducible.

I'm inclined to drop the ~~JuliaLang~~, TestData and ~~Allocation~~ testsets. The former seems problematic from a maintenance standpoint, despite yielding marginal gains [edit: but they barely take up time so they're fine]. TestData and ~~Allocation~~ are now essentially "test tests". All TestData does is check that the functions execute, which will be done implicitly when they are used in the actual test code.

As for the other tests, I'm working on speeding them up by cutting down on fuzzing and possibly introducing some mocking/parallelization.

```
Julia 1.5 - ubuntu-18.04 - x64

Test Summary: | Pass  Total
TestData      |   44     44
 38.361770 seconds (63.49 M allocations: 3.331 GiB, 3.74% gc time)

Test Summary: |  Pass  Total
General       | 10064  10064
 14.073496 seconds (8.47 M allocations: 410.846 MiB, 0.96% gc time)

Test Summary: |  Pass  Total
SurfaceDefs   | 10180  10180
 18.379421 seconds (35.36 M allocations: 1.742 GiB, 3.76% gc time)

Test Summary: |   Pass   Total
Intersection  | 135017  135017
 46.487126 seconds (27.44 M allocations: 1.719 GiB, 1.17% gc time)

Test Summary: |  Pass  Total
Lenses        | 14227  14227
  2.164938 seconds (2.54 M allocations: 130.979 MiB)

Test Summary: |
Emitters      | No tests
  0.002813 seconds (648 allocations: 58.891 KiB)

Test Summary: | Pass  Total
Comparison    |  121    121
 57.904807 seconds (16.88 M allocations: 832.738 MiB, 1.53% gc time)

Test Summary: | Pass  Total
Visualization |    5      5
112.073897 seconds (132.64 M allocations: 6.994 GiB, 2.31% gc time)

Test Summary: | Pass  Total
Allocations   |   26     26
 72.222984 seconds (48.92 M allocations: 2.539 GiB, 66.06% gc time)

Test Summary: | Pass  Total
GlassCat      |  117    117
 55.024593 seconds (69.56 M allocations: 3.497 GiB, 3.26% gc time)
```

alfredclwong:
Would still like to use XUnit.jl to parallelise tests, but the improvements here are already good enough to merit a PR.

Fixes #54

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -38,9 +38,9 @@ jobs:
         run: sudo apt install libxt6 libxrender1 libxext6 libgl1-mesa-glx libqt5widgets5
       - uses: actions/cache@v2
         env:
-          cache-name: cache-artifacts
+          cache-name: cache-julia
         with:
-          path: ~/.julia/artifacts
+          path: ~/.julia/
           key: ${{ runner.os }}-${{ matrix.arch }}-test-${{ env.cache-name }}-${{ hashFiles('**/Project.toml') }}
           restore-keys: |
             ${{ runner.os }}-${{ matrix.arch }}-test-${{ env.cache-name }}-
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -38,9 +38,9 @@ jobs:
         run: sudo apt install libxt6 libxrender1 libxext6 libgl1-mesa-glx libqt5widgets5
       - uses: actions/cache@v2
         env:
-          cache-name: cache-julia
+          cache-name: cache-artifacts
         with:
-          path: ~/.julia/
+          path: ~/.julia/artifacts
           key: ${{ runner.os }}-${{ matrix.arch }}-test-${{ env.cache-name }}-${{ hashFiles('**/Project.toml') }}
           restore-keys: |
             ${{ runner.os }}-${{ matrix.arch }}-test-${{ env.cache-name }}-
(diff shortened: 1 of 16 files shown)


BrianGun:
@alfredclwong Looked at the new runtests code but couldn't tell if the benchmarks tests are run by default every time. We want to keep this otherwise we might introduce a minor change that causes type inference to fail and allocations to increase which will kill performance.  If benchmarks are still run every time then I'll approve

alfredclwong:
Agreed - the Allocations testset is very important. It is included in `alltestsets`:

https://github.com/microsoft/OpticSim.jl/blob/fcca20a96c0fc2dbb4145f27c08c89c301066750/test/runtests.jl#L94

alfredclwong:
I suspect caching all of .julia is causing some issues in 1.5. Will have another look at this tmrw.

alfredclwong:
Caching all of .julia runs us into some weird problems, so I'll just stick with .julia/artifacts. There's probably a good reason why everyone does it this way.

BrianGun (approved):

Issue #86, opened 2021-04-07
Add license text to the top of all source files


BrianGun:
some new source files don't have the MIT license text

alfredclwong:
Sorry, I've been a bit careless with this.

Is it a strong legal requirement? It's been a long time since I've seen full license texts being prepended to every individual source file in a project!

BrianGun:
how about if we go for the second option you have listed above?

alfredclwong:
Yeah I like that one. Maybe something like this.

```
# Copyright (c) Microsoft Corporation. All rights reserved.
# Licensed under the MIT license.
# See LICENSE in the project root for full license information.
```

alfredclwong:
A lot of the files I've added are missing licenses. 😅

Emitters.jl will also be deleted in PR #83.

Next week, I'll try writing a Github Action to automatically create PRs prepending missing licenses to files.

BrianGun:
closing this issue since it is being addressed.

alfredclwong:
I've made a ton of files which are missing licenses across the last month. Correcting this now.

PR #100, opened 2021-04-08
added licence comments to these files.

galran:
following issue #86 added the license comment to these files:
Emitters.jl
NotebooksUtils.jl
Transform.jl


friggog:
The files I added in the Cloud directory are also missing licenses 😬

galran:
> The files I added in the Cloud directory are also missing licenses 😬

got two of them.

PR #101, opened 2021-04-09
(Fake) sequential raytracing support (#63)

### LINKED ISSUE #63: Sequential raytracing
I need sequential raytracing for this package to be useful for my day job, so I would like to take a shot at implementing it.
I would love to have some advice on how to get started with this. I've never before contributed much to an open source project, so I may have some beginner's questions when it comes to the processes for contributing.

cdhf:
Fixes #63. This is not real sequential raytracing, because the system still has to find the correct element the ray hits next. But it allows to select specific beam paths in a complex optical system with eg beamsplitters and makes typical analyses in such systems easier.


BrianGun:
@cdhf thanks for making this change, it will be useful to many people. 

Could you sign the Contributor License Agreement? We can't accept your PR otherwise, it's a Microsoft requirement. If you don't want to sign the CLA then we can submit the PR for you.

BrianGun:
@cdhf have you run the unit tests on your local machine? They are failing on github. Looks like one of the errors occurs at line 186 in processintersection, in Fresnel.jl, called from Comparison.jl in the test sets, somewhere around line 122. 

The error is "opticalinterface is not defined". You can see the errors by clicking on the Details button in the box above this one that says "Some Checks were not successful". You will have to scroll quite a ways to get to the error. 

I believe it's the same error on both ubuntu and windows. 

I looked at that line of code in your repo and don't understand how opticalinterface could be undefined there  since it appears to be defined just above. But something is definitely not right.

Also noticed that you redefined the SphericalLens, etc., functions to take an interfacemode argument. This will set the interfacemode for both front and back surfaces the same. Is this what you wanted?

Does this make sense in general or should we have two interface modes, one for the front and one for the back element, both defaulting to ReflectOrTransmit? I don't have an opinion on this, it's more a matter of what the average (if there is such a thing) optical engineer would find most reasonable and useful.

BrianGun:
@cdhf not sure how experienced you are in Julia. Ignore this if you already know it. To run the unit tests on your local machine type ] at the repl to enter the package mode. Then type test.

cdhf:
It was a typo. I think I forgot to run the tests after changing the conditionals. The tests now work on my machine.

With regard to different interfacemode for front and backside of lenses:
This addresses my more more important use case, so for now it is sufficient for me.

Separate front/backside behaviour makes sense eg to analyse ghost reflections in high power ultra short pulse laser applications. But at least in Zemax I usually had to use this only to work around the bad usability of ray splitting. I sincerely hope that OpticSim will solve that problem at some point. But there is more infrastructure needed for this kind of application, regardless of whether you have raysplitting or not.

BrianGun (approved):

PR #108, opened 2021-04-12
Fix glasscat parsing

### LINKED ISSUE #106: AGF Parsing
The current parsing code for AGF files assumes that lines follow the format:

TOKEN DATA DATA DATA ...
TOKEN DATA DATA DATA ...

However, there are some instances where data are spread over multiple lines, e.g.

```
CD 2.326858000E+000 -2.496168000E-002 2.979231000E-003 3.164392000E-003 
-3.498773000E-004 1.777229000E-005 0.000000000E+000 0.000000000E+000 
0.000000000E+000 0.000000000E+000
```

We need a minor change to the code to accommodate this.

alfredclwong:
Fixes #106

| 1  | 2        | 3         | 4     | 5     | 6                     | 7             | 8         | 9         | 10    | 11    |
|:---|:---------|:----------|:------|:------|:----------------------|:--------------|:----------|:----------|:------|:------|
| NM | raw name | dispform  | ?     |  Nd   | Vd                    | [exclude_sub  | status    | meltfreq] |
| ED | TCE      | ?         | p     |  ΔPgF | [ignore_thermal_exp]  |
| CD | C1       | C2        | C3    |  C4   | C5                    | C6            | C7        | C8        | C9    | C10   |
| TD | D₀       | D₁        | D₂    |  E₀   | E₁                    | λₜₖ           | temp      |
| OD | relcost  | CR        | FR    |  SR   | AR                    | PR
| LD | λmin     | λmax      |
| IT | T1       | T2        | T3    |


alfredclwong:
@BrianGun (or anyone else with the relevant knowledge), could you confirm that the docstring table that I added in c9ec15f is correct for what we want to extract from the AGF files? Some values are ignored - is this intentional?

BrianGun:
@friggog could you verify this code?

friggog:
Looks right, worth noting that most elements of the CD, TD and OD rows are optional I think, the appropriate default values are used currently AFAIK. Also there can be many IT rows which form a table:
| IT | T1       | T2        | T3    |
| IT | T1       | T2        | T3    |
| IT | T1       | T2        | T3    |
| IT | T1       | T2        | T3    |
| IT | T1       | T2        | T3    |
This is used for absorption data lookup/interpolation

This python package might provide some info: https://github.com/nzhagen/zemaxglass - looks like this fork is much more current though: https://github.com/derekjgriffith/zemaxglass

BrianGun:
@friggog @alfredclwong  is the consensus that this PR can be merged and closed?

alfredclwong:
Nothing's really be done yet. I'll write the code up today.

deps/generate.jl
```diff
@@ -71,87 +72,105 @@ function generate_jls(
...
+| IT | T1       | T2        | T3    |
 """
-function sourcefile_to_catalog(source_file::AbstractString)
-    catalog_dict = Dict{String,Dict{String}}()
+function sourcefile_to_catalog(sourcefile::AbstractString)
+    catalogdict = Dict{String,Dict{String}}()
 
-    # check whether the file is UTF8 or UTF16 encoded
-    is_utf8 = isvalid(readuntil(source_file, " "))
-    fo = is_utf8 ? open(source_file, "r") : open(source_file, enc"UTF-16LE", "r")
```
alfredclwong, line 81: @friggog, I'm using DelimitedFiles.readdlm to parse in the .AGF file. Not sure if it'll run into problems with UTF16 encoding - is there a file I can test this on?

alfredclwong: Test fail answered this for me - OHARA.agf :(

Working on a fix

docs/src/examples.md
```diff
@@ -37,6 +37,51 @@ nothing # hide
...
+                  Thickness = [Inf64, 0.0, 5.18, 0.10, 4.40, 0.16, 1.0, 4.96, zoom, 4.04, 1.35, 1.0, 2.80, 3.0, 1.22, dist, missing],
+                  Material = [OpticSim.GlassCat.Air, OpticSim.GlassCat.Air, OpticSim.GlassCat.OHARA.S_LAH66, OpticSim.GlassCat.Air, OpticSim.GlassCat.NIKON.LLF6, OpticSim.GlassCat.Air, OpticSim.GlassCat.OHARA.S_TIH6, OpticSim.GlassCat.OHARA.S_FSL5, OpticSim.GlassCat.Air, OpticSim.GlassCat.OHARA.S_FSL5, OpticSim.GlassCat.Air, OpticSim.GlassCat.OHARA.S_LAL8, OpticSim.GlassCat.OHARA.S_FSL5, OpticSim.GlassCat.Air, OpticSim.GlassCat.OHARA.S_LAH66, OpticSim.GlassCat.Air, missing],
```
alfredclwong, line 65: I changed OHARA.FSL5 -> OHARA.S_FSL5 because none of our default glass catalogs contain the former. Does anyone know if this is innacurate?

alfredclwong: Comparing to previous versions of the example, I can see that we also changed HIKARI.LLF6 to NIKON.LLF6, as well as a few other changes similar to the one I mentioned above. I don't know anything about glass cat naming conventions so it'd be good if someone could confirm that this example is OK.

alfredclwong: https://microsoft.github.io/OpticSim.jl/previews/PR108/examples/#Zoom-Lens

friggog: > I can see that we also changed HIKARI.LLF6 to NIKON.LLF6, as well as a few other changes similar to the one I mentioned above. I don't know anything about glass cat naming conventions

The same material will often be manufactured by a number of companies, the optical properties should be (near) identical, the secondary properties might vary a bit. The results should be near identical, certainly not different enough to matter for an example.

> I changed OHARA.FSL5 to OHARA.S_FSL5 because none of our default glass catalogs contain the former

Companies also come up with new versions of the same material which again have (near) identical optical properties but different secondary properties (cost etc). It should also be safe to substitute these (e.g. BK7 and N_BK7 from SCHOTT) - I'm assuming this is a similar case, but I'm not sure - you can check what differences there are between by looking at the AGF or output of info().

friggog: The zoom lens example is not right - so maybe my assumption about S_FSL5 is wrong - or something else is broken...

alfredclwong: I'll mention it to Brian tonight at our meeting. If we still need to re-optimise the example, I'll take it off this PR so we can merge the important changes for now.

alfredclwong: We'll keep it up for now - PR should be ready for merging, if you're happy with it!

friggog (approved):

PR #111, opened 2021-04-14
[WIP] New feature: add AGF sources

### LINKED ISSUE #107: Robust handling of glass catalog changes
One problem found in #57 is a discontinued glass that seems to be removed in newer versions of the AGF file downloads.

It makes sense to keep even discontinued glasses. Firstly, it can happen that you need to modify an old design to replace a discontinued glass after it has already gone, and then you don't have the old glass to simulate a baseline. Secondly, there are companies that keep stock of old glasses for special applications, so discontinued does not necessarily mean unavailable.

I can think of two ways to do this:

- Seek permission from the glass manufacturers to include the AGF files in this project. The glass catalog is then uniquely identified by the OpticSim version.
- Make easy to use utility functions to store the glass catalogs in the repository of the simulation using this package. The user can then add the files to their own version control system and decide, when/if to update a catalog.

alfredclwong:
Fixes #107

Main commit is e9c1936.

Defines `GlassCat.add_agf(sourcefile::AbstractString; name::Maybe{AbstractString} = nothing, rebuild::Bool = true)`, which allows the user to add custom AGF files to sources.txt at runtime. These will then be incorporated into AGFGlassCat.jl on the next rebuild.

TODO:

- [ ] Write tests for old (and new) GlassCat build functions
- [x] Update docs to reflect changes [[PREVIEW](https://microsoft.github.io/OpticSim.jl/previews/PR111/glasscat/)]
- [ ] Receive feedback


docs/src/glasscat.md
```diff
@@ -1,12 +1,38 @@
...
+the AGF file.
+
+The final two columns are optional, specifying download instructions for acquiring the zipped AGF files
+automatically from the web. The fourth column allows us to use POST requests to acquire files from interactive sites.
+
+When `] build OpticSim` is run, the sources are verified and parsed into corresponding Julia files. These are then
+included in OpticSim via `AGFGlassCat.jl`. These steps are run automatically when the package is first installed using
+`] add OpticSim`, creating a sufficient working environment for our examples and tests.
+
+## Adding glass catalogs
```
BrianGun, line 27: @alfredclwong does this just add a filename to the text file or does it also trigger a build of the AGFGlassCat.jl file?

BrianGun (approved):

BrianGun:
> Will we ever want catalog names that are not purely alphabetical? I can extend the regex to match alphanumeric names with a non-numeric leading character.

for now alphabetical is fine.

PR #143, opened 2021-04-23
[WIP] src/GlassCat/generate.jl tests

alfredclwong:
The existing GlassCat test sets are a little jumbled.

`@testset "Parsing Tests"` parses an AGF file `TEST_CAT.agf` into a Julia Dict `cat`, but then only checks the first glass values.
`@testset "Module Gen Tests"` then checks the generated `AGF_TEST_CAT.jl` against `cat`, which creates a dependency on the previous test set.
`@testset "Glass Tests"` then uses a glass from the `TEST_CAT` module, which also creates a dependency on the previous test set.

Reorganising these and extending the range of checked values to all 4 glasses catches a couple of odd behaviours that have snuck into GlassCat (mostly due to changes that I've made). I'll fix these before opening the PR for review.

### DIFF
--- a/test/testsets/GlassCat.jl
+++ b/test/testsets/GlassCat.jl
@@ -2,32 +2,23 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # See LICENSE in the project root for full license information.
 
+using Test
+using OpticSim
+
+using Unitful
+using Base.Filesystem
+using DataFrames
+using StaticArrays
+
 @testset "GlassCat" begin
     # ensure there aren't any ambiguities
     @test isempty(detect_ambiguities(GlassCat))
     @test isempty(detect_unbound_args(GlassCat))
 
-    ROOT_DIR = joinpath(@__DIR__, "..", "..")
-
-    GLASSCAT_DIR = joinpath(ROOT_DIR, "src", "GlassCat")
-    include(joinpath(GLASSCAT_DIR, "constants.jl"))
-    include(joinpath(GLASSCAT_DIR, "GlassTypes.jl"))
-    include(joinpath(GLASSCAT_DIR, "generate.jl"))
-
-    CATALOG_NAME = "TEST_CAT"
-    SOURCE_DIR = joinpath(ROOT_DIR, "test")
-    SOURCE_FILE = joinpath(SOURCE_DIR, "$(CATALOG_NAME).agf")
-
-    tmpdir = mktempdir()
-    MAIN_FILE = joinpath(tmpdir, "AGF_TEST_CAT.jl")
-    JL_DIR = tmpdir
-
-    cat = Dict()
-
     @testset "Build Tests" begin
         # check that all automatic downloads are working
         for catname in split("HOYA NIKON OHARA SCHOTT Sumita")
-            agffile = joinpath(AGF_DIR, catname * ".agf")
+            agffile = joinpath(OpticSim.GlassCat.AGF_DIR, catname * ".agf")
             @test isfile(agffile)
         end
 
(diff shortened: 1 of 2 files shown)


src/GlassCat/GlassTypes.jl
```diff
@@ -16,13 +16,6 @@ struct GlassID
 end
 
 Base.show(io::IO, a::GlassID) = print(io, "$(string(a.type)):$(a.num)")
-type(a::GlassID) = a.type
-num(a::GlassID) = a.num
-
-ismodel(a::GlassID) = a.type === MODEL
-isMIL(a::GlassID) = a.type === MIL
-isAGF(a::GlassID) = a.type === AGF
-isother(a::GlassID) = a.type === OTHER
```
alfredclwong, line 25: I had a look at some Julia code style discussions online and it seems that getters and setters aren't seen as strictly necessary due to the ability to override Base.getproperty. Any objections?

src/GlassCat/GlassTypes.jl
```diff
@@ -104,16 +97,21 @@ Get the name (including catalog) of the glass, or glass with this ID.
...
-        return OTHER_GLASS_NAMES[num(ID)]
+    elseif t === OTHER
+        return OTHER_GLASS_NAMES[n]
+    elseif t === AGF
+        return AGF_GLASS_NAMES[n]
+    elseif t === TEST
+        return TEST_GLASS_NAMES[n]
     else
-        return AGF_GLASS_NAMES[num(ID)]
+        throw(ArgumentError("unsupported GlassID type $t"))
```
alfredclwong, line 114: I think this is strongly preferable to `else -> AGF`. Similarly for `glassforid()` below

src/GlassCat/generate.jl
```diff
@@ -15,7 +15,11 @@ Each catalog file is a module representing a distinct glass catalog (e.g. NIKON,
 AGF files in `sourcedir`. These are then included and exported in `mainfile`.
 """
 function generate_jls(
-    sourcenames::Vector{<:AbstractString}, mainfile::AbstractString, jldir::AbstractString, sourcedir::AbstractString
+    sourcenames::Vector{<:AbstractString},
+    mainfile::AbstractString,
+    jldir::AbstractString,
+    sourcedir::AbstractString;
+    glasstype::AbstractString = "AGF"
```
alfredclwong, line 22: `glasstype` arg adds support for non-AGF .jl file generation.

test/testsets/GlassCat.jl
```diff
@@ -2,160 +2,128 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # See LICENSE in the project root for full license information.
 
+using Test
+using OpticSim
+
+using Unitful
+using Base.Filesystem
+using DataFrames
+using StaticArrays
```
alfredclwong, line 11: I find this setup quite useful as it allows me to just run `include("test/testsets/GlassCat.jl")`

src/GlassCat/generate.jl
friggog: Is this primarily for testing? I am right in thinking this code only supports generation from AGF files so would always write AGF glass types?

BrianGun:
@friggog and @alfredclwong is this PR ready to merge? A conversation doesn't seem resolved.

src/GlassCat/generate.jl
alfredclwong: Good point. It's probably clearer and safer just to have something like a `test` bool then.

alfredclwong:
@BrianGun I haven't yet resolved Charlie's comment on the new `glasstype` arg. In general I think it's important to have longer, in-depth code review periods and I don't see a need to rush PR merges (unless we're building merge conflicts elsewhere).

BrianGun:
didn't mean to rush anybody. Just wanted to make sure this wasn't forgotten.

src/GlassCat/generate.jl
```diff
@@ -9,14 +9,25 @@ using Unitful
...
 """
 function generate_jls(
-    sourcenames::Vector{<:AbstractString}, mainfile::AbstractString, jldir::AbstractString, sourcedir::AbstractString
+    sourcenames::Vector{<:AbstractString},
+    mainfile::AbstractString,
+    jldir::AbstractString,
+    sourcedir::AbstractString;
+    test::Bool = false
 )
+    glasstype = test ? "TEST" : "AGF"
```
alfredclwong, line 30: Somewhat lazy change but the new feature is hopefully at least a bit clearer now.

friggog: Yeah it's not ideal, but I think preferable to an unconstrained string argument

friggog (dismissed):

friggog (approved):

PR #140, opened 2021-04-22
Switch to new emitter code

alfredclwong:
[Examples preview](https://microsoft.github.io/OpticSim.jl/previews/PR140/examples/)

## Changes
- Update to v0.5.0-DEV (pending detector changes for full v0.5.0 release)
- Remove old Emitter.jl and emitters.md (fixes #127)
  - Rename emitters_new.md to emitters.md
  - Retain RayGenerator types in new file RayGenerator.jl
- Rework examples to use new emitters
  - Split Examples.jl into docs.jl and other.jl to clearly see which examples are used in the docs
  - Use CodeTracking macro @code_string to avoid duplicated code between docs and examples (fixes #144)
- Update tests to use new emitters
- Update Vis to use new emitters
  - Update raygenerator default arg
  - Remove origingen-related draw functions
- Add Maybe and MaybeVector types to OpticSim
- Add a more convenient constructor to Directions.Constant

PR #151, opened 2021-04-28
Fix new emitter switch

alfredclwong:
There are some lingering unresolved review comments and bugs/typos from PR #140. They should be fixed here.


src/maybe.jl
```diff
@@ -0,0 +1,21 @@
...
+"""
+Haskell-style [option type](https://en.wikipedia.org/wiki/Option_type) primarily used to allow optional inputs/outputs
+for functions.
+
+For example, in `src/Examples/docs_examples.jl`, we have `draw_cooketriplet(filename::Maybe{AbstractString} = nothing)`.
+If the function is called with a `filename` argument, then it is passed as a string, and `Vis.save(filename)` saves an
+image of the drawn example. If there is no argument, then the save call is dispatched to the empty `function
+Vis.save(::Nothing) end`. This allows the user to run the example without saving a file.
+"""
+const Maybe{T} = Union{T, Nothing}
```
friggog, line 16: Why doesn't Julia have this by default..?

friggog (approved):

docs/src/examples.md
```diff
@@ -39,7 +39,7 @@ mdparse(@code_string draw_zoomlenses()) # hide
 ```
 
 ```@example example
-syss = draw_zoomlenses(["assets/zoom$i.png" for i in 1:3]); @show sys[1]; nothing # hide
+syss = draw_zoomlenses(["assets/zoom$i.png" for i in 1:3]); @show syss[1]; nothing # hide
```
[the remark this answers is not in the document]
alfredclwong: Could do - admittedly `syss` is an awkward name. I used it just in case we ever want to display more than one of the dataframes.

src/maybe.jl
alfredclwong: In my opinion, it's essentially syntactic sugar, but for some reason it's controversial among the Julia community. There's an official recommendation for `Union{T, Nothing}` [here](https://docs.julialang.org/en/v1/manual/faq/#Nothingness-and-missing-values-1), a long discussion thread [here](https://discourse.julialang.org/t/aliases-for-union-t-nothing-and-union-t-missing/15402/37), and even an example of KristofferC trying to define `const ?{T} = Union{T, Nothing}` [here](https://github.com/JuliaWeb/GitHub.jl/pull/90/files#diff-f3f3eb3e7bb7977f66531d9d1f8855a7R2).

BrianGun: After thinking about it more I'm leaning toward Union{T,Nothing}, without any type aliasing. Just used in place wherever necessary. It's slightly more verbose but extremely clear. 

For people not familiar with Haskell or similar languages Maybe{T} won't mean much - you'll have to look at the definition to figure it out. But Union{T,Nothing} is something any Julia programmer should understand.

In general we should avoid creating type aliases, macros, etc., that make Julia look like a different language. The code will be harder to read and understand for people who aren't part of the inner circle.

PR #152, opened 2021-04-28
Glasscat tests

alfredclwong:
Improve GlassCat test coverage and introduce `wrappedalloc` macro to avoid issues with counting allocations in global scope.


src/GlassCat/GlassTypes.jl
```diff
@@ -338,170 +338,3 @@ function info(io::IO, glass::Glass)
 end
 
 info(g::AbstractGlass) = info(stdout, g)
-
```
BrianGun, line 341: is the docstring function being permanently deleted or did you just move it somewhere? Even though we're not using it much now it seems to be functioning ok. Other people might find it useful for making tools to analyze glasses. Is it necessary to remove it?

alfredclwong: I think I took it out because it seemed like duplicate code (I can't remember where from), it was unused, and it was lowering our test coverage.

friggog: > Just had a look and it seems to be essentially the same as info() above but without the padding. I'll add it back in with a third function to handle with the shared aspects.

This is basically what it did - just printed the same stuff as info with markdown formatting for use in docstrings

alfredclwong: @BrianGun you can have the final say - if you want to keep the code then feel free to `git revert 
d3197a2`

test/testsets/GlassCat.jl
```diff
@@ -117,69 +113,182 @@ using StaticArrays
...
+            "      0.10648μm           0.61        100.0mm",
+            "         0.13μm           0.39        100.0mm",
+            "        0.155μm           0.11        100.0mm",
+            ""
+        ]
+        io = IOBuffer()
+        info(io, CARGILLE.OG0607)
+        @test String(take!(io)) === join(info_lines, '\n')
+
+        # TODO the rest of the string tests
```
friggog, line 264: Should this TODO be addressed now?

test/testsets/GlassCat.jl
```diff
@@ -117,69 +113,182 @@ using StaticArrays
...
+            Sumita,
+        ]
+
+        @test glassnames(CARGILLE) == [
+            :OG0607,
+            :OG0608,
+            :OG081160,
+        ]
+
+        # TODO @test glassnames()
```
friggog, line 283: And this one

test/testsets/GlassCat.jl
```diff
@@ -117,69 +113,182 @@ using StaticArrays
...
+
+        # !! TODO !! for some reason uncommenting this test causes the alloc tests at L145-146 to fail
+        # @test findglass(x -> (x.Nd > 2.1 && x.λmin < 0.5 && x.λmax > 0.9)) == [
+        #     HOYA.E_FDS3,
+        #     Sumita.K_PSFn214P,
+        #     Sumita.K_PSFn214P_M_,
+        # ]
+
+        # TODO _child_modules() unit test
```
friggog, line 292: And this?

test/testsets/GlassCat.jl
```diff
@@ -117,69 +113,182 @@ using StaticArrays
...
+
+        @test glassnames(CARGILLE) == [
+            :OG0607,
+            :OG0608,
+            :OG081160,
+        ]
+
+        # TODO @test glassnames()
+
+        # !! TODO !! for some reason uncommenting this test causes the alloc tests at L145-146 to fail
```
friggog, line 285: This is super weird... Have you tried defining it differently? e.g. define a separate filter function

test/testsets/GlassCat.jl
alfredclwong: Testing `info` is a bit tedious because of all the different possibilities and the fact that it outputs to IO. It doesn't have any dependencies and is probably fine, so I think I'll defer it as a low priority item until another time.

test/testsets/GlassCat.jl
alfredclwong: probably not worth bothering with this at the moment as long as the functions which use this are ok

test/testsets/GlassCat.jl
alfredclwong: quickly writing the test now

test/testsets/GlassCat.jl
alfredclwong: i'll try that out this morning as well as anything else i can think of. might be worth asking for outside help with this one.

alfredclwong: After some digging i've found that the issue is caused by calling a function which takes another function as an argument, such as `function findglass(condition::Function)`. Even if `findglass` does absolutely nothing, this still causes the issue.

We need to fix the allocations test call to run within a local scope to avoid allocations from accessing variables. https://discourse.julialang.org/t/using-allocated-to-track-memory-allocations/4617/5

I'll add the wrappedallocs macro from https://github.com/JuliaAlgebra/TypedPolynomials.jl/blob/master/test/runtests.jl to our runtests.jl

alfredclwong:
@BrianGun I think these changes are ready for approval, if you're happy with them. The wrapped allocations fix is worth a read for future reference.

BrianGun (approved):

PR #154, opened 2021-04-28
get rid of Revise as a dependency

KristofferC:
This seems like an accident. Revise tends to be used by users in case they want to Revise their code but it isn't something packages really need to load.


alfredclwong:
We used to have Revise as an accidental dependency, so I removed it, but then I added it back intentionally to support a feature that relies on reloading source files during runtime.

Essentially, we have a database of glass properties (GlassCat) which consists of a few Julia files containing const definitions. These files are created during Pkg.build and included in OpticSim. There was some demand for a feature where we could add to this database during runtime, and so I decided to use Revise to reload the database after making changes to it.

Do you think this is a reasonable application? I'm open to alternatives as this is becoming an increasingly hacky solution, especially if it is incurring a significant cost in precompilation time.

KristofferC:
> There was some demand for a feature where we could add to this database during runtime, and so I decided to use Revise to reload the database after making changes to it.

I see, I will think a bit about it and see if there might be some other solution that avoids Revise.

alfredclwong:
Thanks! In case you're interested, the relevant code sections are the [build script](https://github.com/microsoft/OpticSim.jl/blob/3dbe82a51fb3c7d2896f19318d3e4756e54fb6d8/deps/build.jl) and [add_agf](https://github.com/microsoft/OpticSim.jl/blob/3dbe82a51fb3c7d2896f19318d3e4756e54fb6d8/src/GlassCat/sources.jl#L12-L50). They both use this [constants file](https://github.com/microsoft/OpticSim.jl/blob/3dbe82a51fb3c7d2896f19318d3e4756e54fb6d8/src/GlassCat/constants.jl).

alfredclwong (changes requested):
block merge for now

BrianGun:
does not seem to be a problem.

PR #171, opened 2021-05-05
Galran/optimization ui1

galran:
Some Major changes in the optimization + ui code.

now supporting the optimization variable syntax of " optional lower bound < VarName=StartValue < optional upper bound " (ex: 10<@=20<30 - @ will create an automatic name for the variable).
lots of changes in the UI part.
zoom in/zoom out in the blink window - Ctrl-+ / Ctrl-- / Ctrl-0 to reset
F12 - toggle dev tools
added support for interactive HTML - using the three_js implementation of WGLMakie to allow real 3D and 2D interaction with plots in the browser.
added the ability to include local assets in the UI, such as images and css files.
added the ability to open/run associated/default program for specific files (for instance, editing in excel or another CSV associated program - should be tested on other OSs)
added the ability to run JS scripts on the shell (not just the browser process)


galran:
@BrianGun, @alfredclwong 

I stuffed quite a lot of different things i wanted to try into the example page - we will need to decide on what to remove and what to leave in.

the two logos at the top are examples of how to use local assets and each one show a different click behavior. the left one will open another Blink window to show the page. The problem is that blink windows do not come with even the basic capabilities of navigating back so it might be good for a single page of documentation etc. the right logo will open a page in the "default" browser - please try it on linux.

i also added a Boolean at the top of the UI source file to allow you to try the interactive HTML - again, needs to be tested on linux. when this flag is true, all the drawings should allow mouse interaction. this is the rreason i had to supply the viewport size in the html because each drawing is not just an image but a webgl based viewport.

please try to run it and ping me with questions, and please don't merge this pr any time soon :)

galran:
of course i forgot to mention that you run it using the following:

OpticSim.Optimization.AxisSymmetricOptimizationUI.run()

alfredclwong:
@BrianGun what is the TODO for here? https://github.com/microsoft/OpticSim.jl/blob/d6df7625192eb5ef03e62c5ab57958113a40d3e7/src/Optimization/OptimizationVariables.jl#L18-L19

friggog:
Would love to see some images of the UI - or are there some in the docs for this PR?

BrianGun:
Excellent question. I wrote this code a long time ago (the recent date in the commit is misleading because Microsoft insisted on a clean commit history when we created the open source repo) so why that TODO is there is no longer clear, if I even put it there. It looks fine to me. 

Could be an @friggog comment - I don't normally use TODO in comments. Maybe he had some notion of future work?

> @BrianGun what is the TODO for here?
> 
> https://github.com/microsoft/OpticSim.jl/blob/d6df7625192eb5ef03e62c5ab57958113a40d3e7/src/Optimization/OptimizationVariables.jl#L18-L19

friggog:
Ahah yeah I think that was mine - just that these bounds are necessarily useful - thickness could be over 100 in theory, zero is maybe a safe lower bound?

alfredclwong:
@galran, going forward could you please try to break up your work into smaller commits? I'm currently reading through the code and it's quite hard to absorb everything at once. For example, the following could have been broken up into separate commits:

1. Displaying the DataFrame, reading CSV files and parsing cell changes
2. Hacking HTML files to replace relative paths with absolute paths
3. Cost function and optimization
4. History slider and Makie displays
5. Misc UI features (zoom, dev tools)
etc.

I know it's sometimes hard to do this while you're figuring stuff out yourself, but then it's just as hard for me to learn everything all at once while reading the code now. ~2000 line code dumps mean that I need to take a day or two just to parse everything. For collaborative work, I think it's worth the extra effort rather than treating Git solely as a tool for uploading code.

src/Optimization/UI/AxisymmetricOptimization.jl
```diff
@@ -0,0 +1,349 @@
...
+using JuMP
+import DataStructures
+
+global global_model = nothing
+global global_ax = nothing
+global global_df = nothing
+global global_counter = 0
+global solution_history = []
+global statistics = Dict()
+global options = Dict()
```
alfredclwong, line 19: Do you think we can put these in a struct that defines an Axisymmetric Optimization problem? I generally feel that global variables are inflexible - this would be more modular, making it easier to implement features such as saving/loading sessions in the future.

src/Optimization/UI/AxisymmetricOptimization.jl
```diff
@@ -0,0 +1,349 @@
...
+    _name::String
+    _start_value::Float64
+    _lower_bound::Union{Float64,Nothing}
+    _upper_bound::Union{Float64,Nothing}
+    _info::Dict
+end
+
+function get_number(dic, key)
+    v = get(dic, key, Nothing)
+    if (v != Nothing)
```
alfredclwong, line 39: n.b. `Nothing` is a type - `nothing` is the actual variable that should be used

it's also more "Julian" to use `isnothing` as this makes use of type dispatch without having to do any comparisons

src/Optimization/UI/AxisymmetricOptimization.jl
```diff
@@ -0,0 +1,349 @@
...
+
+    name = m["name"]
+    start = parse(Float64, m["middle"])
+    lower = m["lower"]===nothing ? nothing : parse(Float64, m["lower"])
+    upper = m["upper"]===nothing ? nothing : parse(Float64, m["upper"])
+
+    var_info = VarInfo(row, col, name, start, lower, upper, Dict())
+
+    return var_info;
+end
```
alfredclwong, line 84: This is an OK solution for now, but going forwards I think that we should take a "UI for dummies" approach where the user doesn't have to learn a specific syntax. Instead, do you think we could have separate columns for min/max/start values?

src/Optimization/UI/AxisymmetricOptimization.jl
```diff
@@ -0,0 +1,349 @@
...
+
+mutable struct VarInfo
+    _row::Int
+    _col::Int
+    _name::String
+    _start_value::Float64
+    _lower_bound::Union{Float64,Nothing}
+    _upper_bound::Union{Float64,Nothing}
+    _info::Dict
+end
```
alfredclwong, line 35: what are the leading underscores meant to mean?

src/Optimization/UI/AxisymmetricOptimization.jl
```diff
@@ -0,0 +1,349 @@
...
+    lower = m["lower"]===nothing ? nothing : parse(Float64, m["lower"])
+    upper = m["upper"]===nothing ? nothing : parse(Float64, m["upper"])
+
+    var_info = VarInfo(row, col, name, start, lower, upper, Dict())
+
+    return var_info;
+end
+
+
+struct Axisymmetric
```
alfredclwong, line 87: This isn't very descriptive. Would AxisymmetricOptimization be more appropriate?

src/Optimization/UI/AxisymmetricOptimization.jl
```diff
@@ -0,0 +1,349 @@
...
+    _variables::AbstractArray
+    
+    function Axisymmetric(def::DataFrame)
+
+        df = copy(def)
+
+        auto_variable_count = 0
+        variables = []
+
+        # first loop - fixing CSV issues and converting items
```
alfredclwong, line 99: I think it would be clearer (more modular) if we split this off into a `csv_to_df` or `parse_csv` function

src/Optimization/UI/AxisymmetricOptimization.jl
```diff
@@ -0,0 +1,349 @@
...
+        end
+            
+        return error
+    catch e
+        @error "Somthing went wrong in calculating the cost function\n$(e)"
+        return 1000 # rqual inf
+    end
+end
+
+function optimize(ax::Axisymmetric, optimization_options::Dict = Dict())
```
alfredclwong, line 279: Are these Ipopt options? I only see :MaxTime begin accessed.

src/Optimization/UI/AxisymmetricOptimizationUI.html
```diff
@@ -0,0 +1,223 @@
...
+}
+.auto-style6 {
+	text-align: center;
+}
+.auto-style7 {
+	text-align: center;
+	font-size: xx-large;
+}
+
+</style>
```
alfredclwong, line 28: We should split this off (and all the other coloring code which is done elsewhere in Julia) into a CSS file.

src/Optimization/UI/AxisymmetricOptimizationUI.jl
```diff
@@ -0,0 +1,546 @@
...
+
+
+function rg_showtable_sync!(w, schema, names, types, rows, coldefs, tablelength, id, options)
+    # RG - disable sorting for all columns
+    for c in coldefs
+        c[:sortable] = false
+    end
+      
+    options[:rowData] = TableView.JSONText(TableView.table2json(schema, rows, types))
+    license = get(ENV, "AG_GRID_LICENSE_KEY", nothing)
```
alfredclwong, line 32: is this license key code relevant?

src/Optimization/UI/AxisymmetricOptimizationUI.jl
```diff
@@ -0,0 +1,546 @@
...
+            return scene, lscene
+        end
+
+        function TableView._showtable_sync!(w, schema, names, types, rows, coldefs, tablelength, id, options)
+            # RG - disable sorting for all columns
+            for c in coldefs
+                c[:sortable] = false
+            end
+            rg_showtable_sync!(w, schema, names, types, rows, coldefs, tablelength, id, options)  
+        end
```
alfredclwong, line 72: Is there no other way to do this other than overwriting a private TableView function? Also you have the code for "RG - disable sorting for all columns" twice (it's also in rg_show_table_sync above). Is this necessary?

src/Optimization/UI/AxisymmetricOptimizationUI.jl
```diff
@@ -0,0 +1,546 @@
...
+import Makie.AbstractPlotting
+import Makie.AbstractPlotting.MakieLayout
+if (INTERACTIVE_HTML)
+    import WGLMakie
+else
+    import GLMakie
+end
+
+
+function rg_showtable_sync!(w, schema, names, types, rows, coldefs, tablelength, id, options)
```
alfredclwong, line 25: i'm not sure that prepending your initials onto function/variable names is a good idea - there's a lot of this in the html file too

src/Optimization/UI/AxisymmetricOptimizationUI.jl
```diff
@@ -0,0 +1,546 @@
...
+        end
+        this.table = @new agGrid.Grid(el, gridOptions)
+        gridOptions.columnApi.autoSizeAllColumns()
+    end
+    onimport(w, handler)
+end
+
+function run()
+
+    @eval begin
```
alfredclwong, line 50: what is this eval block for? runtime redefinition of imported functions?

src/Optimization/UI/AxisymmetricOptimizationUI.jl
```diff
@@ -0,0 +1,546 @@
...
+
+    # this is an optional list of stings to replace inside the html - somthing like an internal macro 
+    # here we are trying to provide "absolute" path to local files relative to the files location
+    a = Base.replace(@__FILE__, "\\" => "/")
+    @show a
+    replace = [                         
+        """src="./""" => """src="$(Base.replace(splitdir(@__FILE__)[1], "\\" => "/"))/""",
+    ]
+
+    html = BlinkUtils.HTMLParserDefault(@__FILE__, replace_in_html = replace)
```
alfredclwong, line 105: is this really necessary?

src/Optimization/UI/BlinkUtils.jl
```diff
@@ -0,0 +1,352 @@
...
+    _text
+    _variables
+end
+
+function HTMLParser(filename::String; kwargs...)
+    sl = readlines(filename)
+    return BlinkUtils.HTMLParser(sl; kwargs...)
+end
+
+function HTMLParser(text::AbstractArray{String}; replace_in_html = nothing, kwargs...)
```
alfredclwong, line 50: I'm not keen on the idea of pre-parsing html files before passing them to Blink. It seems to add an unnecessary layer of complexity to the development.

Are relative paths not workable (the logos actually aren't even displaying on my machine)? Are these comment-based hooks necessary when you can (presumably, correct me if I'm wrong) use queryselector?

If we take this idea further, we'll almost be developing our own HTML Julia wrapper.

src/Optimization/UI/AxisymmetricOptimization.jl
```diff
@@ -0,0 +1,349 @@
...
+    upper = m["upper"]===nothing ? nothing : parse(Float64, m["upper"])
+
+    var_info = VarInfo(row, col, name, start, lower, upper, Dict())
+
+    return var_info;
+end
+
+
+struct Axisymmetric
+    _original_definition
```
alfredclwong, line 88: what is this?

src/Optimization/UI/AxisymmetricOptimization.jl
BrianGun: Julia Computing tech support told us that we should use

``` v === nothing```
and 
```v !== nothing```

instead of
```isnothing(v)```

They said the compiler can do more extensive optimization with the === and !== statements.

alfredclwong: Interesting, thanks for the correction. I was looking at this link and didn't read all the way to the end. I'll have to correct some of my code then. https://stackoverflow.com/questions/56852880/comparing-julia-variable-to-nothing-using-or

alfredclwong:
I'm ok to merge these changes into the optimization branch if you are @galran @BrianGun. The review comments I left above were more questions/discussion points than changes that I want to make immediately.

BrianGun:
this is a scratchpad branch for now so it can be rougher than usual.

PR #179, opened 2021-05-10
Allow for surface type extensibility in AxisymmetricOpticalSystem

### LINKED ISSUE #176: Support more surface types in AxisymmetricOpticalSystem
Currently we only support `:Object`, `:Image`, `:Stop` as surface types in our Axisymmetric Optical System. Would be good to get at least Zernike surfaces supported. This would require an extension to the DataFrame structure to support a large number of generic coefficients/parameters.

alfredclwong:
Fixes #176

I'll add support for Zernike surfaces in another PR - thought it'd be best to keep this one focused on reworking the core framework first.

## Changes

1. Changes to the prescription DataFrame:
    - Rename `Surface` column header to `SurfaceType`
    - Use Strings instead of Symbols in the `SurfaceType` column (e.g. `:Object` -> `"Object"`)
    - Add a `Parameter` column to allow for surface type extensibility
2. Support Standard and Aspheric surfaces
3. Update examples, sample notebooks, diagnostics, test data and Eye.jl to use new scheme
4. Write a function to validate prescription DataFrames
5. ~~[TODO] Write tests for AxisymmetricOpticalSystem~~ (will probably rework the system soon so not worth it at the moment)


src/Optical/OpticalSystem.jl
```diff
@@ -231,7 +231,7 @@ function validate_axisymmetricopticalsystem_dataframe(prescription::DataFrame)
         "Material" => GlassCat.AbstractGlass,
         "SemiDiameter" => Real
     ]
-    supported_surface_types = ["Object", "Stop", "Image", "Standard", "Conic", "Aspheric", "Zernike"]
+    supported_surface_types = ["Object", "Stop", "Image", "Standard", "Aspheric", "Zernike"]
```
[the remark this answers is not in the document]
friggog: In Zemax it is, we can do it as we like, probably simplest include within Standard

src/Examples/docs_examples.jl
```diff
@@ -40,11 +40,16 @@ function draw_zoomlenses(filenames::Vector{<:Union{Nothing,AbstractString}} = re
...
+    ]
     syss = [
         AxisymmetricOpticalSystem{Float64}(DataFrame(
             SurfaceType = ["Object", "Stop", "Standard", "Standard", "Standard", "Aspheric", "Standard", "Standard", "Aspheric", "Aspheric", "Standard", "Standard", "Standard", "Standard", "Standard", "Standard", "Image"],
             Radius = [Inf64, Inf64, -1.6202203499676E+01, -4.8875855327468E+01, 1.5666614444619E+01, -4.2955326460481E+01, 1.0869565217391E+02, 2.3623907394283E+01, -1.6059097478722E+01, -4.2553191489362E+02, -3.5435861091425E+01, -1.4146272457208E+01, -2.5125628140704E+02, -2.2502250225023E+01, -1.0583130489999E+01, -4.4444444444444E+01, Inf64],
-            Parameters = [missing, missing, missing, missing, missing, [(4, 1.0386E-04), (6, 1.4209E-07), (8, -8.8495E-09), (10, 1.2477E-10), (12, -1.0367E-12), (14, 3.6556E-15)], missing, missing, [(4, 4.2721E-05), (6, 1.2484E-07), (8, 9.7079E-09), (10, -1.8444E-10), (12, 1.8644E-12), (14, -7.7975E-15)], [(4, 1.1339E-04), (6, 4.8165E-07), (8, 1.8778E-08), (10, -5.7571E-10), (12, 8.9994E-12), (14, -4.6768E-14)], missing, missing, missing, missing, missing, missing, missing],
```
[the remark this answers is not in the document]
friggog: Not a big fan of this, firstly we only support even aspherics so either we need to double the index, or every other index will always be 0. Secondaly, it is might be the case that only high order aspherics are used so for a single parameter (12th aspheric term) we need a list of length 24. The same arguments are true for most surface types with many parameters eg zernike, qtype, chebyshev. This gets even worse where we have multiple parameters eg zernike surface which has both zernike terms and aspherics, how do we input these in a flat list format?

Is it not possible to support arbitrarily typed parameters, or even a dictionary of param_name: param_value so that not all parameters need to be provided for a surface type which has multiple parameters. This makes input _much_ clearer to the user as well, otherwise they have to somehow parse a flat list which might potentially represent a number of different parameters

src/Examples/docs_examples.jl
```diff
@@ -41,9 +41,9 @@ function draw_zoomlenses(filenames::Vector{<:Union{Nothing,AbstractString}} = re
     raygenerator = Sources.Source(; transform, origins, directions)
 
     aspherics = [
-        [0, 0, 0, 1.0386E-04, 0, 1.4209E-07, 0, -8.8495E-09, 0, 1.2477E-10, 0, -1.0367E-12, 0, 3.6556E-15],
-        [0, 0, 0, 4.2721E-05, 0, 1.2484E-07, 0, 9.7079E-09, 0, -1.8444E-10, 0, 1.8644E-12, 0, -7.7975E-15],
-        [0, 0, 0, 1.1339E-04, 0, 4.8165E-07, 0, 1.8778E-08, 0, -5.7571E-10, 0, 8.9994E-12, 0, -4.6768E-14],
+        [4 => 1.0386E-04, 6 => 1.4209E-07, 8 => -8.8495E-09, 10 => 1.2477E-10, 12 => -1.0367E-12, 14 => 3.6556E-15],
+        [4 => 4.2721E-05, 6 => 1.2484E-07, 8 => 9.7079E-09, 10 => -1.8444E-10, 12 => 1.8644E-12, 14 => -7.7975E-15],
+        [4 => 1.1339E-04, 6 => 4.8165E-07, 8 => 1.8778E-08, 10 => -5.7571E-10, 12 => 8.9994E-12, 14 => -4.6768E-14],
```
alfredclwong, line 46: Not quite a dictionary, but I think Pairs here are more expressive than tuples regarding our intended usage. This will easily convert to a Dict if needed later.

src/Examples/other_examples.jl
```diff
@@ -37,9 +37,9 @@ function cooketriplet(::Type{T} = Float64, detpix::Int = 1000) where {T<:Real}
         DataFrame(
             SurfaceType = ["Object", "Standard", "Standard", "Standard", "Stop", "Standard", "Standard", "Image"],
             Radius = [Inf, 26.777, 66.604, -35.571, 35.571, 35.571, -26.777, Inf],
-            OptimizeRadius = [false, true, true, true, true, true, true, false],
+            # OptimizeRadius = [false, true, true, true, true, true, true, false],
```
alfredclwong: These columns can be reintroduced later when the Optimization code is more mature. But they shouldn't be used within the AxisymmetricOpticalSystem constructor. Instead, we should have some kind of AxisymmetricOptimization constructor which parses the extra columns.

Project.toml
```diff
@@ -2,7 +2,7 @@ name = "OpticSim"
 uuid = "24114763-4efb-45e7-af0e-cde916beb153"
 authors = ["Brian Guenter", "Charlie Hewitt", "Ran Gal", "Alfred Wong"]
 repository = "https://github.com/microsoft/OpticSim.jl"
-version = "0.5.0"
+version = "0.5.0-DEV"
```
alfredclwong: We still need a bug fix from PR #183 before releasing v0.5.0.

alfredclwong (commented):
@friggog @BrianGun this PR should be ready for merging now

src/Examples/docs_examples.jl
friggog: Yeah I think there is a better way to do this still eg where surface type is the actual type name allowing you to get the constructor for the surface dynamically and then you could pass the parameter dict as keyword args. This is a pretty fundamental change and might mean modifying the surface constructors too (ughh) but would mean that new surface types would be supported directly and there wouldn’t be any need for hard coding stuff when setting up axisymmetric systems. 

Anyway this seems reasonable for now, but I think there is a better solution there that might be easiest to achieve by starting almost from scratch on the design of AxisymmetricOpticalSystem

alfredclwong: I see what you mean - got the wrong idea earlier!

It'll definitely require a lot of changes, and it'll get mixed up with some other discussions we're having now about prescription DataFrames for non-axisymmetric systems... (although this also means it's a good time for a rewrite). Let's go with this for now - in the short term, all I want is to add support for Zernike surfaces. I'll factor this into our discussions and work towards something smoother in the future.

BrianGun (approved):

PR #182, opened 2021-05-11
[WIP] Lenses.jl rework

alfredclwong:
Opening a draft PR to discuss existing code using the review mechanism.

### DIFF
--- a/src/Optical/Lenses.jl
+++ b/src/Optical/Lenses.jl
@@ -2,123 +2,376 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # See LICENSE in the project root for full license information.
 
-opticinterface(::Type{S}, insidematerial, outsidematerial, reflectance = zero(S), interfacemode = ReflectOrTransmit) where {S<:Real} = FresnelInterface{S}(insidematerial, outsidematerial, reflectance = reflectance, transmission = one(S) - reflectance, interfacemode = interfacemode)
+using .OpticSim.GlassCat: AbstractGlass, Air
+
+export SphericalLens, ConicLens, AsphericLens, FresnelLens
+
+function opticinterface(
+    ::Type{T},
+    insidematerial::AbstractGlass,
+    outsidematerial::AbstractGlass,
+    reflectance::T = zero(T),
+    interfacemode::InterfaceMode = ReflectOrTransmit
+) where {T<:Real}
+    transmission = one(T) - reflectance
+    return FresnelInterface{T}(insidematerial, outsidematerial; reflectance, transmission, interfacemode)
+end
 
 """
-    SphericalLens(insidematerial, frontvertex, frontradius, backradius, thickness, semidiameter;  lastmaterial = OpticSim.GlassCat.Air, nextmaterial = OpticSim.GlassCat.Air, frontsurfacereflectance = 0.0, backsurfacereflectance = 0.0, frontdecenter = (0, 0), backdecenter = (0, 0), interfacemode = ReflectOrTransmit)
+    SphericalLens(
+        insidematerial,
+        frontvertex,
+        frontradius,
+        backradius,
+        thickness,
+        semidiameter;
+        lastmaterial = Air,
+        nextmaterial = Air,
+        frontsurfacereflectance = 0.0,
+        backsurfacereflectance = 0.0,
+        frontdecenter = (0, 0),
+        backdecenter = (0, 0),
+        interfacemode = ReflectOrTransmit
+    )
 
 Constructs a simple cylindrical lens with spherical front and back surfaces. The side walls of the lens are absorbing.
 """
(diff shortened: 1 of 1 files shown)


src/Optical/Lenses.jl
```diff
@@ -2,123 +2,374 @@
...
+        frontsurfacereflectance = 0.0,
+        backsurfacereflectance = 0.0,
+        frontdecenter = (0, 0),
+        backdecenter = (0, 0),
+        interfacemode = ReflectOrTransmit
+    )
 
 Constructs a simple cylindrical lens with spherical front and back surfaces. The side walls of the lens are absorbing.
 """
-function SphericalLens(insidematerial::T, frontvertex::S, frontradius::S, backradius::S, thickness::S, semidiameter::S; lastmaterial::Q = OpticSim.GlassCat.Air, nextmaterial::R = OpticSim.GlassCat.Air, frontsurfacereflectance::S = zero(S), backsurfacereflectance::S = zero(S), frontdecenter::Tuple{S,S} = (zero(S), zero(S)), backdecenter::Tuple{S,S} = (zero(S), zero(S)), interfacemode = ReflectOrTransmit) where {R<:OpticSim.GlassCat.AbstractGlass,Q<:OpticSim.GlassCat.AbstractGlass,T<:OpticSim.GlassCat.AbstractGlass,S<:Real}
```
alfredclwong, line 12: Was there any reason for parameterising `AbstractGlass` types `R`, `Q`, `T`? This was done throughout the file.

src/Optical/Lenses.jl
```diff
@@ -2,123 +2,376 @@
...
+    nsamples::Int = 17,
+    frontdecenter::Tuple{T,T} = (zero(T), zero(T)),
+    backdecenter::Tuple{T,T} = (zero(T), zero(T)),
+    interfacemode::InterfaceMode = ReflectOrTransmit
+) where {T<:Real}
+    @assert semidiameter > zero(T)
     @assert !isnan(frontradius)
 
-    frontdecenter_l = abs(frontdecenter[1]) > eps(S) && abs(frontdecenter[2]) > eps(S) ? sqrt(frontdecenter[1]^2 + frontdecenter[2]^2) : zero(S)
-    backdecenter_l = abs(backdecenter[1]) > eps(S) && abs(backdecenter[2]) > eps(S) ? sqrt(backdecenter[1]^2 + backdecenter[2]^2) : zero(S)
```
alfredclwong, line 43: Is it possible to do `x != zero(S)` instead of `abs(x) > eps(S)` here? Not sure if there's a particular use case of decentering that we are accounting for.

src/Optical/Lenses.jl
```diff
@@ -2,123 +2,376 @@
...
+        zero(T)
+    )
+    backdecenter_l = (
+        abs(backdecenter[1]) > eps(T) && abs(backdecenter[2]) > eps(T) ?
+        sqrt(backdecenter[1]^2 + backdecenter[2]^2) :
+        zero(T)
+    )
+
+    # front lens
+    frontinterface = opticinterface(T, insidematerial, lastmaterial, frontsurfacereflectance, interfacemode)
```
alfredclwong, line 200: Other than interface, is there any reason to treat the front and back lenses separately? I'm reading and comparing both halves of the code to spot the differences now.

src/Optical/Lenses.jl
```diff
@@ -2,123 +2,376 @@
...
+        lens_rear = leaf(
+            surf,
+            Transform{T}(zero(T), convert(T, π), zero(T), backdecenter[1], backdecenter[2], frontvertex - thickness)
+        )
     end
-    extra_front = frontradius >= zero(S) || isinf(frontradius) ? zero(S) : abs(frontradius) - sqrt(frontradius^2 - semidiameter^2)
-    extra_back = backradius >= zero(S) || isinf(backradius) ? zero(S) : abs(backradius) - sqrt(backradius^2 - semidiameter^2)
+
+    extra_front = (
+        frontradius >= zero(T) || isinf(frontradius) ?
```
alfredclwong, line 322: Does the first condition make the second condition redundant?

src/Optical/Lenses.jl
friggog: Don't think so

src/Optical/Lenses.jl
friggog: I think eps was used to allow for some numeric error as the decenter values are calculated, not sure if there is much practical impact to use zero instead - maybe add a test for this edge case?

src/Optical/Lenses.jl
friggog: Front and back could have different reflectance - in the future they may have totally different interfaces due to different coatings being present on each side (very common)

alfredclwong: Vague wording from me here - I was trying to get an understanding of how much of the somewhat duplicated code is unique between the front/back lenses. My motivation for this is that I'll probably be splitting the lens code into

1. Various functions for creating a (front or back) lens surface of each surface type
2. A function for combining each of these surfaces into a CSG lens

I wanted to double-check that the only significant differences between front and back would be their interfaces and a z-translation. Is this correct?

PR #183, opened 2021-05-12
Replace lingering generateray() calls

alfredclwong:
There were some lingering calls to `generateray`, which is from the old Emitters API, in the `trace` function. It seems that the same functionality can be achieved with the new API by doing the following replacement:

`generateray(sources, k)` -> `sources[k]`

@galran could you confirm?

On a side note, this means that none of our multithreaded trace functions are being tested (confirmed here https://codecov.io/gh/microsoft/OpticSim.jl/src/main/src/Optical/OpticalSystem.jl#L376). @BrianGun is this because it was particularly difficult to test multithreaded code, or did we just skip it?

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -2,7 +2,7 @@ name = "OpticSim"
 uuid = "24114763-4efb-45e7-af0e-cde916beb153"
 authors = ["Brian Guenter", "Charlie Hewitt", "Ran Gal", "Alfred Wong"]
 repository = "https://github.com/microsoft/OpticSim.jl"
-version = "0.5.0-DEV"
+version = "0.5.0"
 
 [deps]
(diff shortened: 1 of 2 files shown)


BrianGun:
@alfredclwong this was an oversight. Don't think the multithreaded code is that hard to test, at least to make sure that the code doesn't crash when you run it. 

> On a side note, this means that none of our multithreaded trace functions are being tested (confirmed here https://codecov.io/gh/microsoft/OpticSim.jl/src/main/src/Optical/OpticalSystem.jl#L376). @BrianGun is this because it was particularly difficult to test multithreaded code, or did we just skip it?

BrianGun (approved):

PR #187, opened 2021-05-13
Fix type validation and parsing in Parameter column

alfredclwong:
1. Make Parameters be subtypes of `Vector{<:Pair{<:AbstractString,<:Real}}` which allows for flexibility beyond numerically indexed coefficients.
2. Fix the parsing of Aspherics to convert this into a `Vector{Tuple{Int,S}}` to conform with existing surface constructors (temporary fix, we'll probably rewrite the constructors soon).

Currently (without these changes), `Examples.draw_zoomlenses` doesn't work.


src/Optical/OpticalSystem.jl
```diff
@@ -374,7 +374,12 @@ struct AxisymmetricOpticalSystem{T,C<:CSGOpticalSystem{T}} <: AbstractOpticalSys
             elseif surface_type == "Aspheric"
                 semidiameter = convert(T, max(get_front_back_property(prescription, i, "SemiDiameter", zero(T))...))
```
BrianGun, line 375: don't think a call to convert is necessary here; pretty sure it is called implicitly since semidiameter is declared of type T.

BrianGun (commented):
It's cleaner not to have the explicit convert(T...), or convert(Vector{T}...) statements. What do you think?

src/Optical/OpticalSystem.jl
alfredclwong: Agreed, this is much better. I was using `convert(T, x)` to replace several usages of `T(x)` which were being highlighted as "possible method call errors" by VSCode.

I'm adding a commit that makes better use of type declarations. This also helps with readability.

src/Optical/OpticalSystem.jl
```diff
@@ -393,7 +387,7 @@ struct AxisymmetricOpticalSystem{T,C<:CSGOpticalSystem{T}} <: AbstractOpticalSys
                 )
             end
 
-            if firstelement && semidiameter !== NaN
+            if firstelement
                 systemsemidiameter = semidiameter
                 firstelement = false
             end
```
alfredclwong, line 393: This change will allow the Stop semidiameter to define the systemsemidiameter if it's the first element. This is different from the previous behaviour - is this OK?

src/Optical/OpticalSystem.jl
```diff
@@ -180,7 +181,7 @@ function trace(
             # this will almost always not apply as the detector will be in air, but it's possible that the detector is
             # not in air, in which case this is necessary
             if !isair(m)
-                mat = glassforid(m)::OpticSim.GlassCat.Glass
+                mat::Glass = glassforid(m)
```
alfredclwong, line 184: Is `glassforid(m)::OpticSim.GlassCat.Glass` equivalent to `convert(OpticSim.GlassCat.Glass, glassforid(m)`, or is it more of a type assertion?

Either way, in this case, am I right to say that the type declaration `mat::Glass` is preferable?

BrianGun:
```
mat = glassforid(m)::OpticSim.GlassCat.Glass
```
says that mat will be assigned something of type Glass but does not fix the type of mat.
```
mat::Glass = glassforid(m)
```
fixes the type of mat. The latter allows the compiler to make more assumptions about mat, since it's type is not fixed and cannot change anywhere else in the program. Probably better to do it this way in general.

src/Optical/OpticalSystem.jl
BrianGun: ask Joel Kollin. He could tell us if this makes sense optically.

alfredclwong: I asked the Cambridge team today and the general consensus was that individual semidiameters are important but not so much the system semidiameter. And, if anything, it'd be more useful to have this as the max value. I'll have a look around the code tomorrow to see if this is actually used anywhere currently.

BrianGun (approved):

PR #188, opened 2021-05-17
Infix notation for CSG operations

alfredclwong:
This allows us to do things like

`barrel ∩ topsurface ∩ botsurface`

instead of

`csgintersection(barrel, csgintersection(topsurface, botsurface))`

I've kept the old functions for backwards compatibility and the ability to apply intermediate transforms. We could also rethink this approach, though.


src/Geometry/CSG/CSG.jl
```diff
@@ -126,12 +128,14 @@ Create a (pseudo) leaf node from another CSGGenerator, this is useful if you wan
 function leaf(n::CSGGenerator{T}, transform::Transform{T} = identitytransform(T))::CSGGenerator{T} where {T<:Real}
     return CSGGenerator{T}((parenttransform) -> n(parenttransform * transform))
 end
-export leaf
 
 # CSG objects are trees, not graphs, since it makes no sense to reuse intermediate results. Hence no need to keep track of common subexpressions.
 # Static arrays are much faster than mutable arrays so want node transforms to be static. Unfortunately the transformations cascade from the root to the leaves but the nodes have to be created from the leaves to the roots.
 # Wrap the csg operations in function that delays evaluation until the transform has been computed.
 
+!(a::CSGGenerator{T}) where {T<:Real} = CSGGenerator{T}((parenttransform) -> ComplementNode(a(parenttransform)))
```
friggog, line 136: We don't want to expose this as it will allow users to construct CSG trees that may not conform the assumption we make for bounding box culling of rays - users should only use intersection, union and difference

friggog:
> I've kept the old functions for backwards compatibility and the ability to apply intermediate transforms. We could also rethink this approach, though.

Given that you can create a leaf node solely to apply a transform I think that the intermediate transforms are now unnecessary. Using leaves to do this is slightly less concise in terms of number of function calls required, but should evaluate to the same thing in the end. It possibly also makes it simpler for the user as there is only one way to apply a transform - using a leaf node, though if the are going to be used a lot within the tree maybe we should rename it to transform or something 

Broadly I'm a big fan of this change - much less typing and much more readable! 
I think we should probably figure out the transforms properly before merging it though (and deprecate the old method completely too? removing exports will likely break any downstream code anyway)

src/Geometry/CSG/CSG.jl
alfredclwong: good point, i'll remove it

alfredclwong:
I'd be more than happy to deprecate the old notation - will have a discussion with @BrianGun and @galran to see their thoughts.

> Given that you can create a leaf node solely to apply a transform I think that the intermediate transforms are now unnecessary. Using leaves to do this is slightly less concise in terms of number of function calls required, but should evaluate to the same thing in the end. It possibly also makes it simpler for the user as there is only one way to apply a transform - using a leaf node, though if the are going to be used a lot within the tree maybe we should rename it to transform or something

I had exactly the same minor concern about the verbosity of intermediate transforms, which is what I meant by "rethinking the approach". It's probably fine - as you say, it's a little cleaner in terms of user interface - but open to discussion.

> removing exports will likely break any downstream code anyway)

The exports are actually moved to the top, not removed. Personally I think this is better code style for a maturing project as it more clearly shows the API that a particular source file is defining.

BrianGun:
What are you proposing? Eliminating intermediate transformation nodes or using ∩, ∪, - instead of csgunion, etc. Or both?

No objection to overloading ∩, ∪, -. If you do this then define another function transform(a::CSGTree,b::Transform). This would create an intermediate transformation node. Then you would do this
```
Sphere() ∪ transform(A,transformation)
```
instead of what we currently do:
```
csgunion(Sphere(),A,transformation)
```

Eliminating intermediate transformation nodes doesn't seem like a good idea. Forcing the user to specify transformations only at the leaves would complicate the construction of geometric hierarchies. Where are those leaf node transformations going to be computed? 

The user would have to maintain a separate geometric transformation tree that matches the csg tree and then to extract transformations from the leaves of the geometric transformation tree and insert them into the constructors of the matching leaf nodes in the csg tree. That seems complicated.

In our current system the examples have shallow hierarchies but it's not hard to imagine systems with more complex hierarchies. It seems simpler to include the transformations along with the csg tree

alfredclwong:
I don't think anyone's thinking of eliminating intermediate transforms.

Originally, I just wanted to add ∩, ∪, - for convenience, on top of the existing functions, but after converting a few examples I think this is a much better notation to use for CSG operations and would support removing the old functionality.

Currently I think `leaf` does exactly what we'd want `transform` to do. As Charlie suggested, we can probably just repurpose it for the new interface. I'll look into this tomorrow.

BrianGun:
leaf should create a LeafNode, a node with no children, although I see that there is a new leaf function that does something different.

It's confusing to overload leaf to make it mean something else. If we change the second definition of leaf to transform then that will make the naming more consistent with function.

src/Geometry/CSG/CSG.jl
```diff
@@ -115,69 +125,94 @@ end
...
 end
 """
     leaf(surf::CSGGenerator{T}, transform::Transform{T} = identitytransform(T)) -> CSGGenerator{T}
 
-Create a (pseudo) leaf node from another CSGGenerator, this is useful if you want multiple copies of a premade CSG structure with different transforms, for example in an MLA.
+Create a (pseudo) leaf node from another CSGGenerator, this is useful if you want multiple copies of a premade CSG
+structure with different transforms, for example in an MLA.
 """
-function leaf(n::CSGGenerator{T}, transform::Transform{T} = identitytransform(T))::CSGGenerator{T} where {T<:Real}
+function leaf(n::CSGGenerator{T}, transform::Transform{T} = identitytransform(T)) where {T<:Real}
```
friggog, line 137: yeah exactly as Brian suggests:
```suggestion
function transform(n::CSGGenerator{T}, transform::Transform{T}) where {T<:Real}
```

src/Geometry/CSG/CSG.jl
```diff
@@ -2,14 +2,18 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # See LICENSE in the project root for full license information.
 
+import Base: ∩, ∪, -
+
+export CSGTree, CSGGenerator, leaf
```
friggog, line 7:
```suggestion
export CSGTree, CSGGenerator, leaf, transform
```

alfredclwong:
I've had a proper look at the CSG code now - I was a bit confused by the pseudo-leaf yesterday as I didn't take the time to question what it was actually doing. I agree that `transform` would be a much more appropriate name for it.

However, this may have been indicative of the fact that the underlying system was a little awkward to work with in the first place. I find it somewhat unintuitive that the `CSGTree` isn't actually the object that we interface with during CSG tree creation. Rather, we create a tree wrapper of sorts using various `CSGGenerator` function chains, mirroring paths in the tree. There are three ways to alter this transform tree:

1. A leaf node can be created with an initial transform (returns `CSGGenerator`).
2. Each binary node (union, intersection, difference) can be created whilst simultaneously creating a transform to propagate down the tree at evaluation (returns `CSGGenerator`).
3. A final transform can be applied to the root object before evaluation (returns `CSGTree`).

In my opinion, this is a peculiar design. If intermediate transforms are indeed vital to the process, then why not incorporate them into the `CSGTree` structure? Each node can easily store a transform alongside its children. This makes applying transforms at any point very simple: just change the transform field of the corresponding node. There's also the added bonus of making inverse computations a little bit more efficient as they can be evaluated using the tree structure.

I'll try making these changes to show what I mean. Feel free to disagree/point out any flaws in this proposed design.

BrianGun:
There were several things driving the current design. The most important was that we wanted to minimize the cost of transforming rays. We also wanted to be able to make reusable subsystems where the entire subsystem could be transformed after it was created. 

To properly transform the ray we only need the final leaf transform, which is the composition of all the transforms on the path from the root to the leaf. But this is a static array for efficiency so it can't be changed after the subsystem is created. That's a problem if the creation of the subsystem and the transformation of the subsystem are in different parts of the code, if the subsystem is in a library of premade cparts for example.

The current system addresses this problem by making a subsystem factory which has the intermediate transforms encoded as arguments to function closures. To create a new subsystem you execute the factory, optionally passing in a root node transform. Then it constructs the subsystem by executing the function closures, storing the resulting arrays only at the leaf nodes.

An alternative would be to traverse the csg tree from the root and make a copy as you go. When you reach a leafnode you'd multiply the transform stored there by the incoming transform and use that as the argument to the LeafNode constructor to make the new leaf node.

This is exactly what the current code does. It's simple and efficient and only takes 15 lines of code, 12 lines of which are simple multidispatch method definitions. If you rewrote it you'd end up with code that does exactly the same thing, with equivalent or greater complexity. 

Since the current code is not broken and the replacement code would have exactly the same functionality it's not clear that it's worth rewriting. It's a different style of coding using closures that is more common in Julia. Perhaps it needs to be better documented so it is easier to understand. 

We should stop exporting the csgintersection,csgunion,csgdifference functions and instead export the standard ∩, ∪, - infix operations. The end programmer would write a  ∩ transform(b) instead of the current csgintersection(a,b,transform).

alfredclwong:
In general, I don't have a problem with using function closures to delay the evaluation of the leaf node transforms. The current solution is coded nicely for this design choice.

However, I'm trying to consider it from a data structure standpoint. Especially since we're moving towards more interactive applications, we may want to be able to view and edit these intermediate transforms. In my opinion, having these values stored within functions makes them harder to access compared to if they were just fields in a mutable node.

I think that my suggestion ([CSGTree2](https://github.com/microsoft/OpticSim.jl/blob/515025e1b5b7f6a235f6c2679895d02fe005d336/src/Geometry/CSG/CSG2.jl)) simplifies both the underlying implementation and exposed interface by collapsing everything back into the CSGTree, including the transforms (maintaining the ability to delay evaluation and apply intermediate transforms) and different types of nodes (by adding another parameterisation to the struct).

I wrote a [demo script](https://github.com/microsoft/OpticSim.jl/blob/515025e1b5b7f6a235f6c2679895d02fe005d336/src/Geometry/CSG/demo.jl) to show an example use case. On top of the existing functionality, it is also possible to print the CSG tree, outputting something like this:

```
OpticSim.CSGIntersection: transform = [-1.0 0.0 1.2246467991473532e-16 0.0; 0.0 1.0 0.0 0.0; -1.2246467991473532e-16 -0.0 -1.0 -5.0; 0.0 0.0 0.0 1.0]
├─ OpticSim.CSGIntersection: transform = I
│  ├─ OpticSim.CSGLeaf: transform = I, surface = Cylinder{Float64}(9.0, FresnelInterface(SCHOTT.N_BK7, GlassCat.Air, 0.0, 0.0, ReflectOrTransmit))  
│  └─ OpticSim.CSGLeaf: transform = [1.0 0.0 0.0 0.0; 0.0 1.0 0.0 0.0; 0.0 0.0 1.0 5.0; 0.0 0.0 0.0 1.0], surface = AcceleratedQTypeSurface{Float64, 3, 3, 2}(9.0, -0.04, 0.3, Cylinder{Float64}(9.0, FresnelInterface(GlassCat.Air, GlassCat.Air, 0.0, 0.0, ReflectOrTransmit)), [0.0, 0.0], SVector{2, Float64}[[-1.538089935299395, 2.138089935299395], [0.0, 0.0], [0.0, 0.0]], SVector{2, Float64}[[-0.2, 0.0], [0.565685424949238, 0.0], [-0.6531972647421808, 0.0]], 9.5, 2.4000000000000004)
│
└─ OpticSim.CSGLeaf: transform = I, surface = Plane{Float64}([0.0, 0.0, -5.0], [0.0, 0.0, -1.0], FresnelInterface(SCHOTT.N_BK7, GlassCat.Air, 0.0, 1.0, ReflectOrTransmit))
```

src/Geometry/CSG/CSG2.jl
```diff
@@ -0,0 +1,83 @@
...
+"""
+    CSGTree{N<:CSGType,T<:Real}
+
+A binary tree representing a CSG structure. `N` defines the root node's type: Leaf, Union, Intersection or Difference.
+
+The leaf nodes of a `CSGTree` contain parametric surfaces. Non-leaf nodes define CSG operations between their children.
+At each node, it is possible to store a transform that propagates down the tree at evaluation.
+"""
+mutable struct CSGTree2{N<:CSGType,T<:Real} <: Primitive{T}
+    data::CSGData{T}
```
friggog, line 50: These are all abstract/unsized types so we're going to have a real bad time with allocations as it is.
I think you'll need both left and right type parameterized, also the CSGData wrapper will probably need the surface type parameterized.

friggog:
> Especially since we're moving towards more interactive applications, we may want to be able to view and edit these intermediate transforms. In my opinion, having these values stored within functions makes them harder to access compared to if they were just fields in a mutable node.

This is a very good point, but in Julia mutable == bad performance (pretty much) - if we want to go this route it might be best to work on changing CSGGenerator to be mutable, but still have it 'compile' to a CSGTree which is immutable and fully typed to retain good performance at trace time.

alfredclwong:
> This is a very good point, but in Julia mutable == bad performance (pretty much) - if we want to go this route it might be best to work on changing CSGGenerator to be mutable/directly modifiable, but still have it 'compile' to a CSGTree which is immutable and fully typed to retain good performance at trace time.

Writing high performance code in Julia is something I'm still not completely up to scratch on, so thanks for the pointers. Forgot to think about this when I was criticising the current design.

_If_ we could have everything done by one struct, it seems that we would want an immutable tree that we can construct from the leaves up, with (parent and) child pointers, and the ability to edit data during and after construction at each node. Given these constraints, I think you're right - we'll either need to rebuild the entire tree for each change or follow some kind of 'compilation' flow, which would be an adaptation on CSGGenerator. It might actually be simplest to repurpose "CSGTree2" for this and write a `compile(::CSGTree2)::CSGTree` function.

BrianGun:
@alfredclwong  Interactive modification of transformations might be a good reason to make the more extensive changes you propose. For sure make the mods to switch to infix.

Does your code rewrite for csgtransformation that incorporates mutable transformations work? Does it have a compile function that generates statically typed immutable data structures? If you've already done all this work and all test pass then we could switch to the new representation.

But if you haven't done all this then it would be good to examine the other tasks on our list. Getting the UI for optimization working better is high on that list. Do you think this task is more important, or essential before working on the UI?

-brian

alfredclwong:
> @alfredclwong Interactive modification of transformations might be a good reason to make the more extensive changes you propose. For sure make the mods to switch to infix.
> 
> Does your code rewrite for csgtransformation that incorporates mutable transformations work? Does it have a compile function that generates statically typed immutable data structures? If you've already done all this work and all test pass then we could switch to the new representation.
> 
> But if you haven't done all this then it would be good to examine the other tasks on our list. Getting the UI for optimization working better is high on that list. Do you think this task is more important, or essential before working on the UI?
> 
> -brian

It'd probably take another day or two to get a full rewrite done, but I agree it's not the highest on our priority list at the moment. The previous commit was more to open a discussion on possible future ideas.

I'll revert back to the minimal infix change, make the `leaf->transform` change, and I'll also keep the old functions (csgunion etc.) with a deprecation warning in order to avoid making breaking changes.

alfredclwong:
@BrianGun this should be ready for review now. We can go through it briefly today at our meeting.

BrianGun (approved):

PR #195, opened 2021-05-24
Replace leaf/transform with csgtransform

alfredclwong:
After using the new interface a little bit, I don't think it's a good idea to have both `leaf` and `transform` available for applying transforms to CSG objects. Also `transform` is a popular name, both for variables and functions.

I'd suggest keeping `leaf` as a private method for converting a ParametricSurface into a CSGGenerator (without transforming), and changing `transform` to `csgtransform`, able to handle both ParametricSurface and CSGGenerator types.

### DIFF
--- a/docs/src/csg.md
+++ b/docs/src/csg.md
@@ -12,7 +12,7 @@ The code for this in our system would look this this:
 ```@example
 using OpticSim # hide
 cyl = Cylinder(0.7)
-cyl_cross = cyl ∪ leaf(cyl, Geometry.rotationd(90, 0, 0)) ∪ leaf(cyl, Geometry.rotationd(0, 90, 0))
+cyl_cross = cyl ∪ csgtransform(cyl, Geometry.rotationd(90, 0, 0)) ∪ csgtransform(cyl, Geometry.rotationd(0, 90, 0))
 
 cube = Cuboid(1.0, 1.0, 1.0)
 sph = Sphere(1.3)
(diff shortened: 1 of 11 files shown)


BrianGun:
@alfredclwong Agree with taking the transform out of the leaf constructor. However, why is it be better to name the function csgtransform than transform? 

transform is used in many places but it has the same meaning here as it does when applied to points or geometry. Giving it a different name implies that its function is different somehow for csg objects which is confusing. 

If you transform(point), transform(geometry), etc. wherever we already use transform() then naturally you're going to expect to transform(csgobject). And it will be irritatingly inconsistent if you have to use a function with a different name.

alfredclwong:
I agree - `transform` would be ideal. Perhaps I'm being lazy, but there were two inconveniences:

- Namespace clash with `DataFrames.transform` means you either have to explicitly import OpticSim: transform or use OpticSim.transform
- As far as I know, having a method called transform means that you can't also use that symbol for a variable in the same scope

Together this probably means that we have to type out OpticSim.transform for every usage of the function, which is a bit annoying. Is there a way around this?

BrianGun:
@alfredclwong what functions do we use from dataframe? Could we import dataframe instead of using it, and just take the functions we need? I'm pretty sure we don't use the DataFrame.transform function. That would get rid of the name clash.

BrianGun:
@alfredclwong ```OpticSim.jl``` has this
```
using DataFrames: DataFrame
```
but ```Examples.jl``` has this
```
using DataFrames
```
I believe that the ```DataFrame``` constructor is the only function we use from ```DataFrames```. 

Examples.jl is included in ```OpticSim.jl``` so we could remove the ```using DataFrames``` statement from ```Examples.jl``` and that should eliminate the name clash between ```DataFrames.transform``` and our ```transform()``` function defined for CSG objects.

BrianGun:
We fixed the name clash with DataFrames.transform in another PR so using transform seems more sensible than using csgtransform. closing this PR.

PR #208, opened 2021-06-24
Poor performance of ray generation in Emitters.jl because of abstract struct field

BrianGun:
## Description

Fixes #  207 Poor performance of ray generation in 

Changed Integer to Int64 in struct field declarations in Emitters code. The previous declaration was causing allocations and slow execution.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## How Has This Been Tested?

ran unit tests

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules
- [ ] I have checked my code and corrected any misspellings


Project.toml
```diff
@@ -5,6 +5,7 @@ repository = "https://github.com/microsoft/OpticSim.jl"
 version = "0.5.1"
 
 [deps]
+BenchmarkTools = "6e4b80f9-dd63-53aa-95a3-0cdb28fa8baf"
```
friggog, line 8: I don't think BenchmarkTools should be a dependency of the package itself - it's already [installed separately for tests](https://github.com/microsoft/OpticSim.jl/blob/9eec4c489c958847c55c83c01c559b614ab87b64/test/Project.toml#L2)

src/Examples/other_examples.jl
```diff
@@ -254,34 +254,6 @@ function fresnel(convex = true; kwargs...)
     Vis.drawtracerays(sys; test = true, trackallrays = true, numdivisions = 30, kwargs...)
 end
 
-function grating(; period = 1.0, θ = 0.0, λ = 0.55, kwargs...)
```
friggog, line 257: These removals seem unrelated to the PR - is this intentional? If so maybe should be a separate PR

test/TestData/TestData.jl
```diff
@@ -10,12 +10,33 @@ module TestData
...
 using Unitful
 using Images
 using Base: @.
 using LinearAlgebra
+using BenchmarkTools
+
+
+### Emitters
+
+function RectGridAllocations()
```
friggog, line 25: 100% agree on adding a test for Emitter allocations to ensure this kind of regression doesn't happen again - but it should really go in [Benchmarks.jl](https://github.com/microsoft/OpticSim.jl/blob/main/test/Benchmarks/Benchmarks.jl) and be added to [`all_benchmarks`](https://github.com/microsoft/OpticSim.jl/blob/9eec4c489c958847c55c83c01c559b614ab87b64/test/Benchmarks/Benchmarks.jl#L67) which will then be tested as part of the [allocations tests](https://github.com/microsoft/OpticSim.jl/blob/main/test/testsets/Allocations.jl)

test/testsets/Emitters.jl
```diff
@@ -6,6 +6,7 @@ using Test
 using OpticSim
 using OpticSim.Emitters
 using OpticSim.Geometry
+using BenchmarkTools
```
friggog, line 9: I don't think this is needed

galran (dismissed):

Project.toml
BrianGun: See comment below.

src/Examples/other_examples.jl
BrianGun: It was intentional. A separate PR would be better in an ideal world but because it was a deletion of code that didn't work anymore it was simpler to add it into a single PR.

test/testsets/Emitters.jl
BrianGun: yeah, that was an accident. I put it in temporarily because it was much faster to verify that the fixes were working than running the test set.

This is a weakness of our current setup. There are many times when you need to run a quick test like this, when you are first debugging the code. If this is done in the test sets then it's a 5 minute cycle as opposed to a few seconds if BenchmarkTools is a dependency. 

I'll remove the BenchmarkTools dependency for now.

galran (approved):

PR #209, opened 2021-06-25
Multidetectors

### LINKED ISSUE #141: Do not ignore detector in the middle of an optical system
A detector can only be hit if there is no other element that can be hit.

Consider this system:

```
w1 = leaf(Cuboid(20.0, 20.0, 2.0; 
                            interface = FresnelInterface{Float64}(SCHOTT.N_BK7, Air; 
                                                                                       reflectance=0.0, transmission=1.0)),
                translation(0.0, 0.0, -30.0))
w2 = leaf(Cuboid(20.0, 20.0, 2.0; 
                            interface = FresnelInterface{Float64}(SCHOTT.N_BK7, Air; 
                                                                                        reflectance=0.0, transmission=1.0)),
                translation(0.0, 0.0, -60.0))
la = LensAssembly(w1(), w2())
detector = Rectangle(20.0, 20.0, SVector(0.0, 0.0, 1.0), SVector(0.0, 0.0, -50.0); interface = opaqueinterface())
sys = CSGOpticalSystem(la, detector)
Vis.drawtracerays(sys(); trackallrays=true, rayfilter=nothing, colorbynhits=true)
Vis.drawtraceimage(sys())
```
The drawtracerays visualization shows rays passing through the detector and hitting the lower windown. The drawtraceimage visualization stays dark. Tracing a single ray will return nothing. This seems to happen because the trace function in OpticalSystem only considers intersections with the detector after the ray has ben traced through all elements of the lens assembly. At this time the rays are already below the detector and propagating in the wrong direction.

This behavior is not intuitive for me. There are also use cases where one needs to place a detector at an intermediate position. 

How I would like it to work:
- The detector is treated like a normal surface during raytracing of the lens assembly
- If its interface is opaque, the ray is terminated at the detector
- Otherwise the ray can continue to propagate through the system, possibly hitting other detectors (once OpticSim has this capability)
- A ray is terminated, if it hits an opaque detector, if its power drops below a user defined threshold, or if the recursion depth is too large

cdhf:
WIP to fix #141


src/Geometry/Surface.jl
```diff
@@ -27,6 +27,7 @@ Basic `Surface`s are _not_ valid CSG objects, they function only in a stand-alon
 ```julia
 surfaceintersection(surface::Surface{T}, ray::AbstractRay{T,3}) -> Union{EmptyInterval{T},Interval{T}}
 normal(surface::Surface{T}) -> SVector{3,T}
+surface_id(surface::Surface{T}) -> UUID
```
BrianGun, line 30: We could define interface and surface_id once in Surface.jl like this:

surface_id(surf::Surface) = surf.surface_id
interface(surf::Surface) = surf.interface

Then all surfaces would inherit these functions and we wouldn't have to replicate them. A few surfaces need special surface_id and interface functions such as Rectangle

surface_id(surf::Surface) = surface_id(surf.plane)

but that should work fine. If a function defines a more specialized surface_id or interface function that's the one that will be executed rather than the default defined in Surface.jl.

Not essential to make this change but would simplify the code a bit.

src/Optical/Detectors.jl
```diff
@@ -0,0 +1,73 @@
+# MIT license
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# See LICENSE in the project root for full license information.
+
+import Base
+
+"""
+    AbstractDetector{T<:Number}
```
BrianGun, line 8: Not sure we need the type in AbstractDetector. For some other cases having the parametric types in the abstract type was important to give the compiler enough information to generate good code. Not sure that will be necessary here.

src/Optical/LensAssembly.jl
```diff
@@ -324,14 +324,16 @@ Recursive rays are offset by a small amount (`RAY_OFFSET`) to prevent it from im
 
 `trackrays` can be passed an empty vector to accumulate the `LensTrace` objects at each intersection of `ray` with a surface in the assembly.
 """
-function trace(assembly::LensAssembly{T}, r::OpticalRay{T,N}, temperature::T = T(OpticSim.GlassCat.TEMP_REF), pressure::T = T(OpticSim.GlassCat.PRESSURE_REF); trackrays::Union{Nothing,Vector{LensTrace{T,N}}} = nothing, test::Bool = false, recursion::Int = 0)::Union{Nothing,NoPower,LensTrace{T,N}} where {T<:Real,N}
+function trace(system, assembly::LensAssembly{T}, r::OpticalRay{T,N}, temperature::T = T(OpticSim.GlassCat.TEMP_REF), pressure::T = T(OpticSim.GlassCat.PRESSURE_REF); trackrays::Union{Nothing,Vector{LensTrace{T,N}}} = nothing, test::Bool = false, recursion::Int = 0)::Union{Nothing,NoPower,LensTrace{T,N}} where {T<:Real,N}
+    detector_id = surface_id(detector(system))
```
BrianGun, line 328: do we want to allow for multiple detectors? Maybe detector_id::AbstractVector{T} where{T<:AbstractDetector}.

src/Optical/LensAssembly.jl
```diff
@@ -353,8 +355,11 @@ function trace(assembly::LensAssembly{T}, r::OpticalRay{T,N}, temperature::T = T
                 # i.e. with the path length for this intersection added and power modulated by absorption only
                 push!(trackrays, LensTrace(OpticalRay(ray(r), raypower, λ, opl = raypathlength, nhits = nhits(r), sourcenum = sourcenum(r), sourcepower = sourcepower(r)), intsct))
             end
+            if intersection_id == detector_id
```
BrianGun, line 358: if we allow for multiple detectors this would be
```
if in(intersection_id, detector_ids)
```

src/Optical/LensAssembly.jl
```diff
@@ -353,8 +355,11 @@ function trace(assembly::LensAssembly{T}, r::OpticalRay{T,N}, temperature::T = T
...
+                update!(detectorimage(system), detector(system), LensTrace(OpticalRay(ray(r), raypower, λ, opl = raypathlength, nhits = nhits(r), sourcenum = sourcenum(r), sourcepower = sourcepower(r)), intsct))
```
BrianGun, line 359: multiple calls to update here maybe?

src/Optical/OpticalSystem.jl
```diff
@@ -45,14 +45,13 @@ CSGOpticalSystem(
...
-    detector::S
-    detectorimage::HierarchicalImage{D}
+    detector::Pair{S, AbstractDetector{D}}
     temperature::T
     pressure::T
 
     function CSGOpticalSystem(
         assembly::L,
-        detector::S,
+        detector::Pair{S, String},
```
BrianGun, line 54: Don't understand what this constructor would be used for.

src/Optical/OpticalSystem.jl
```diff
@@ -62,17 +61,29 @@ struct CSGOpticalSystem{T,D<:Number,S<:Surface{T},L<:LensAssembly{T}} <: Abstrac
...
+function CSGOpticalSystem(
+    assembly::L,
+    detector::S,
+    detectorpixelsx::Int = 1000,
+    detectorpixelsy::Int = 1000,
+    t::Type{D} = Float32;
+    temperature::Union{T,Unitful.Temperature} = convert(T, TEMP_REF),
+    pressure::T = convert(T, PRESSURE_REF)
+    ) where {T<:Real,S<:Surface{T},L<:LensAssembly{T},D<:Number}
+    CSGOpticalSystem(assembly, detector => "Default Detector", detectorpixelsx, detectorpixelsy, t; temperature = temperature, pressure = pressure)
```
BrianGun, line 84: Again, don't understand what the string is for.

src/Optical/OpticalSystem.jl
```diff
@@ -147,79 +158,7 @@ function trace(
         return nothing
     end
 
-    result = trace(system.assembly, r, temperature(system), pressure(system), trackrays = trackrays, test = test)
```
BrianGun, line 150: Not understanding how you can delete this. It's not the same as the trace in LensAssembly. It does the memory managment for the interval pools, for example.

src/Optical/LensAssembly.jl
cdhf: Definitely, haven't done this yet, but multiple detectors are definitely useful.

src/Optical/OpticalSystem.jl
cdhf: To give detectors a name, once multiple detectors are implemented, so you can refer to detectors by name later, or use the name for reporting

cdhf:
I am sorry, but for the time being I don't have the time to finish this. I've taken on another role at work, which currently doesn't leave me time for much else. Should I close the pull request?

BrianGun:
Cdhf,

No reason to apologize. We're grateful for the contributions you did make. And we are currently implementing something very much like this. Don't close the pull request yet, we're using the discussion there as a reference. It's better to leave it open as a reminder to us to keep working on it.

Have fun in your new role. I hope you have the opportunity to contribute to OpticSim again. Your comments were always spot on and your suggestions were technically excellent.

-brian

BrianGun:
No progress on this PR for a long time so closing it.

PR #213, opened 2021-07-01
Added a new "Surface" type: ConvexPolygon which support planner simple convex polygons

galran:
added two new libraries dependencies - one is part of the Julia basic libs, the other one is for fast processing of 2D polygons.

Added a new "Surface" child: ConvexPolygon which support planner simple convex polygons. These can be used to create non-symmetric paraxial lenses.
Moved a couple of utility functions form the Emitters module to the Geometry module where they belong. These include right, up, forward and origins for type Transform.
Also added some tests for the new ConvexPolygon type.


src/Geometry/Primitives/NonCSG/ConvexPolygon.jl
```diff
@@ -0,0 +1,104 @@
...
+ConvexPolygon(local_frame::Transform{T}, local_polygon_points::Vector{Vector{T}}, interface::NullOrFresnel{T} = nullinterface(T))
+```
+
+The local frame defines the plane (spans by the right and up vectors) with the plane normal given by the forward vector.
+the local_polygon_points are given with respect to the local frame and are 2D points.
+"""
+struct ConvexPolygon{T} <: Surface{T}
+    plane::Plane{T,3}
+    local_frame::Transform{T}
+    local_points::Vector{Vector{T}}
```
BrianGun, line 25: local_points should probably be SVector{SVector}. Vectors will be allocated on the heap which will lead to allocations and will probably also be slower. If we expected large convex polygons then Vector might be appropriate. But this class is going to be used for polygons with less than 20-30 points. If you change to SVector probably should document that should only have small number of points.

src/Geometry/Primitives/NonCSG/ConvexPolygon.jl
```diff
@@ -0,0 +1,104 @@
...
+    _poly2d::GeometricalPredicates.Polygon2D{GeometricalPredicates.Point2D}
+
+    function ConvexPolygon(
+            local_frame::Transform{T},
+            local_polygon_points::Vector{Vector{T}}, 
+            interface::NullOrFresnel{T} = NullInterface(T)
+        ) where {T<:Real}
+
+        @assert length(local_polygon_points) > 3         # need at least 3 points to define apolygon
+        @assert size(local_polygon_points[1])[1] == 2    # check that first point is a 2D point
```
BrianGun, line 37: if you use SVector instead of Vector then you can specify vector size variable declaration and can eliminate this assert.

src/Geometry/Primitives/NonCSG/ConvexPolygon.jl
```diff
@@ -0,0 +1,104 @@
...
+        new{T}(plane, local_frame, local_polygon_points, poly)
+    end
+end
+export ConvexPolygon
+
+# Base.show(io::IO, poly::ConvexPolygon{T}) where {T<:Real} = print(io, "ConvexPolygon{$T}($(centroid(hex)), $(normal(hex)), $(hex.side_length), $(interface(hex)))")
+centroid(poly::ConvexPolygon{T}) where {T<:Real} = poly.plane.pointonplane
+interface(poly::ConvexPolygon{T}) where {T<:Real} = interface(poly.plane)
+normal(poly::ConvexPolygon{T}) where {T<:Real} = normal(poly.plane)
+normal(poly::ConvexPolygon{T}, ::T, ::T) where {T<:Real} = normal(poly)
```
BrianGun, line 55: what is this function for? If you really need two then the reason should probably be in a documentation string for the normal function.

src/Geometry/Primitives/NonCSG/ConvexPolygon.jl
```diff
@@ -0,0 +1,104 @@
...
+function surfaceintersection(poly::ConvexPolygon{T}, r::AbstractRay{T,3}) where {T<:Real}
+    interval = surfaceintersection(poly.plane, r)
+    if interval isa EmptyInterval{T} || isinfiniteinterval(interval)
+        return EmptyInterval(T) # no ray plane intersection or inside plane but no hit
+    else
+        intersect = halfspaceintersection(interval)
+        p = point(intersect)
+
+        local_p = world2local(poly.local_frame) * p
+        @assert abs(local_p[3]) < 0.00001   # need to find a general epsilon - check that the point lies on the plain
```
BrianGun, line 67: Not sure this assert is necessary. If we have a ray/plane intersection then local_p should always be a valid point. An absolute test is subject to problems with scale, if the convex polygons become large for some reason. Could lead to erroneous failures.

src/Geometry/Primitives/NonCSG/ConvexPolygon.jl
```diff
@@ -0,0 +1,104 @@
...
+function makemesh(poly::ConvexPolygon{T}, ::Int = 0) where {T<:Real}
+    c = centroid(poly)
+
+    l2w = local2world(poly.local_frame)
+    l = length(poly.local_points)
+
+    triangles = []
+    for i in 1:l
+        p1 = poly.local_points[i]
+        p2 = poly.local_points[mod(i,l) + 1]
```
BrianGun, line 92: might want to use a variable name other than l here since it looks almost exactly the same as 1 in the code font. Maybe len instead of l?

src/Geometry/Primitives/NonCSG/ConvexPolygon.jl
```diff
@@ -0,0 +1,104 @@
...
+                return positivehalfspace(intersect)
+            else
+                return rayorigininterval(intersect)
+            end
+        end
+    end
+end
+
+
+function makemesh(poly::ConvexPolygon{T}, ::Int = 0) where {T<:Real}
```
BrianGun, line 83: is makemesh used for drawing the convex poly in 3D? Maybe a documentation string as to the purpose of this function.

src/Geometry/Transform.jl
```diff
@@ -233,6 +234,15 @@ function Transform(rotation::AbstractArray{T,2}, translation::AbstractArray{T,1}
         translation[1], translation[2], translation[3], one(T))
 end
 
+
+# define some utility functions 
+right(t::Transform{<:Real}) = normalize(Vec3(t[1,1], t[2,1], t[3,1]))
```
BrianGun, line 239: documentation strings would be helpful here to explain what these functions do and what they are used for. Graphics people will know but optics people might not.

src/utilities.jl
```diff
@@ -71,3 +71,7 @@ end
         return acos(x)
     end
 end
+
+
+# some place holders for package level function names
+function origin end
```
BrianGun, line 77: documentation string would be useful otherwise few people will understand what this function definition is for. Maybe show an example of overloading this function?

src/Geometry/Primitives/NonCSG/ConvexPolygon.jl
galran: I replaced it with Vector{SVector{2, T}}.
I can't find a convenient way to define an static array of static arrays. Please send me an example if you have one.

src/Geometry/Primitives/NonCSG/ConvexPolygon.jl
galran: I copied it from the Hexagon example. I think that its just a matter of interface support.
You can have a normal for the surface and a normal for each point on the surface in the case when it's not planer.

galran: It seems that all the shapes have the same interface - i think that normal without parameters is the one that should be removed but on the other hand what point would you give to a shape that doesn't have a natural parametrization.

BrianGun:
Depends on what you mean by convenient. SVector{N,SVector{2,T}} where N is whatever the number of 2vectors is. Is this what you were asking or did I misunderstand?

galran:
> Depends on what you mean by convenient. SVector{N,SVector{2,T}} where N is whatever the number of 2vectors is. Is this what you were asking or did I misunderstand?

I tried it and could not make it work:
# -----------------------------------------------------
using StaticArrays
function func(points::SVector{N,SVector{2,T}})  where {N<:Int64, T<:Real}
    println(points)
end
v = SVector(SVector(1.0, 1.0), SVector(2.0, 2.0), SVector(3.0, 3.0))
func(v)
# -----------------------------------------------------

resulting in the error:
ERROR: LoadError: MethodError: no method matching func(::SVector{3, SVector{2, Float64}})
Closest candidates are:
  func(::SArray{Tuple{N}, SVector{2, T}, 1, N}) where {N<:Int64, T<:Real} at d:\Projects\Rays\Github\galran\Polygons\Try5.jl:4
Stacktrace:
 [1] top-level scope
   @ d:\Projects\Rays\Github\galran\Polygons\Try5.jl:10
in expression starting at d:\Projects\Rays\Github\galran\Polygons\Try5.jl:10

what am i doing wrong here?

src/Geometry/Primitives/NonCSG/ConvexPolygon.jl
BrianGun: you'd have to do something like this:

struct ConvexPolygon{N,T<:Real} <: Surface{T}
        local_points::SVector{N,SVector{2,T}}

or maybe even better

struct ConvexPolygon{Npoints,Dim,T<:Real} <: Surface{T}
        local_points::SVector{Npoints,SVector{Dim,T}}

where Dim is the dimension of the points, and Npoints is the number of points.

BrianGun: @rgal does my suggestion for defined SVector{SVector} work for this case?

galran: I'm getting the following error when trying it:
ERROR: LoadError: TypeError: in ConvexPolygon, in N, expected N<:Int64, got a value of type Int64.
this is my entire test:
using StaticArrays

struct ConvexPolygon{N<:Int64, T<:Real} 
    local_points::SVector{N, SVector{2, T}}

    function ConvexPolygon(points::SVector{N, SVector{2, T}}) where {N<:Int64, T<:Real}
        new{N, T}(points)
    end

end

points = SVector(SVector(1.0, 1.0), SVector(2.0, 2.0), SVector(3.0, 3.0))
p = ConvexPolygon{3, Float64}(points)

BrianGun: You don't want this:
```
struct ConvexPolygon{N<:Int64,T<:Real}
```
you want this:
```
struct ConvexPolygon{N,T<:Real}
```
The first statement is saying you want N to be a type which inherits from Int64, which is the same as saying a type equal to Int64 because there are no subtypes of Int64. But the argument you will use in the constructor is a number, not a type. 

you'll want to get rid of the N<:Int64 in the ConvexPolygonn constructor as well. I made those two changes and the code worked. This should do it:

```
struct ConvexPolygon{N, T<:Real} 
    local_points::SVector{N, SVector{2, T}}

    function ConvexPolygon(points::SVector{N, SVector{2, T}}) where {N, T<:Real}
        new{N, T}(points)
    end

end
```

src/Geometry/Primitives/NonCSG/ConvexPolygon.jl
```diff
@@ -13,50 +14,62 @@ The rotation of the polygon around its normal is defined by `rotationvec`.
...
 end
 export ConvexPolygon
 
 # Base.show(io::IO, poly::ConvexPolygon{T}) where {T<:Real} = print(io, "ConvexPolygon{$T}($(centroid(hex)), $(normal(hex)), $(hex.side_length), $(interface(hex)))")
-centroid(poly::ConvexPolygon{T}) where {T<:Real} = poly.plane.pointonplane
-interface(poly::ConvexPolygon{T}) where {T<:Real} = interface(poly.plane)
-normal(poly::ConvexPolygon{T}) where {T<:Real} = normal(poly.plane)
-normal(poly::ConvexPolygon{T}, ::T, ::T) where {T<:Real} = normal(poly)
-
+centroid(poly::ConvexPolygon{N, T}) where {N, T<:Real} = poly.plane.pointonplane
```
BrianGun, line 67: Since N and T are not being used on the right hand side of the = you shouldn't need them. Should still generate efficient code if you remove them.

src/Optical/Paraxial.jl
```diff
@@ -17,10 +17,10 @@ ParaxialLensConvexPoly(focaldistance, local_frame, local_polygon_points, local_c
 ```
 """
 struct ParaxialLens{T} <: Surface{T}
-    shape::Union{Rectangle{T},Ellipse{T},Hexagon{T},ConvexPolygon{T}}
```
BrianGun: I don't think this will do what you want. I think this will define a ConvexPolygon whose first type argument is T. This might be what you want:

```
shape::Union{Rectangle{T},Ellipse{T},Hexagon{T},ConvexPolygon{N,T}} where{N}
```

Then for sure the T type argument to ConvexPolygon is in the right place.

BrianGun (commented):
@galran  I think the ConvexPolygon{T} thing will cause problems since that puts the T type argument in the first place, where the N argument should go.

BrianGun (approved):

PR #215, opened 2021-07-03
latticeExtensions

BrianGun:
## Description

Adds Repeat module with functions for repeated structures such as pixel grids.

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## How Has This Been Tested?

not yet tested. Only for feedback

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules
- [ ] I have checked my code and corrected any misspellings


src/RepeatingStructures/HexagonalLattice.jl
```diff
@@ -0,0 +1,129 @@
...
+
+"""Returns ```centerpoint``` plus all its neighbors in a ring of size n. This is a hexagonal region of the lattice n times as large as the unit hexagon cell.
+
+Example:
+
+```
+Vis.@wrapluxor Vis.drawhexcells(50, hexregion((0,0),2))
+```
+"""
+function hexregion(centerpoint::Tuple{Int64,Int64}, n::Int64)
```
galran, line 55: is HexBasis supposed to be a parameter to this function?

src/RepeatingStructures/HexagonalLattice.jl
```diff
@@ -0,0 +1,129 @@
...
+```
+"""
+function hexregion(centerpoint::Tuple{Int64,Int64}, n::Int64) 
+    f(i) = i==0 ? () : (neighbors(HexBasis1,centerpoint,i)...,f(i-1)...)
+    return (centerpoint,f(n)...)
+end
+export hexregion
+
+"""Returns all hex cells contained in the rings up to size n centered around centerpoint. Does not include centerpoint. Use hexregion for that."""
+function neighbors(::Type{HexBasis1}, centerpoint::Tuple{Int64,Int64},n) where{T}
```
galran, line 62: Seems like this function returns the list of neighbors at exactly n distance from the centerpoint, and not "up to". I think that is the region function.

galran (approved):
This is some big chunk of code.
The API looks good to me.
Maybe you can add an example of some simple version of RGB pixels to demonstrate a more complex "cell"

BrianGun:
@cdhf if you have time could you look over and comment on this api for creating repeating structures? You've had good insights in the past, would appreciate the input.

src/Examples/Examples.jl
```diff
@@ -19,6 +20,7 @@ using LinearAlgebra
 
 include("docs_examples.jl")
 include("other_examples.jl")
+include("ExamplesRepeatingStructures.jl")
```
friggog, line 23: should follow the naming convention ie `repeating_structure_examples.jl`

docs/make.jl
```diff
@@ -19,6 +19,7 @@ makedocs(
             "Basic Types" => "basic_types.md",
             "Primitives" => "primitives.md",
             "CSG" => "csg.md"
```
friggog, line 21:
```suggestion
            "CSG" => "csg.md",
```

src/Examples/Examples.jl
BrianGun: what was the reasoning behind this? It's clearly what we are doing but it seems strange that everywhere else we capitalize and do not use _ but in the examples files we use all lowercase and _. Is there some good reason for doing this? Or maybe we should make the example files follow the naming convention of the rest of the repo.

friggog: The convention is for files containing modules to be capitalised and for ‘scripts’ to be lowercase with underscores

src/RepeatingStructures/HexagonalLattice.jl
BrianGun: good idea. then we have region and neighbors, which work for any basis.

src/RepeatingStructures/HexagonalLattice.jl
BrianGun: You are correct.

galran (approved):
for some reason the documentation is not finishing which is weird.

BrianGun:
Yeah, I noticed that too. Earlier I let it run for over 2 hours and it didn't finish. Could it be the latest changes we made to get the illustrations to display, maybe putting Documenter.jl in an infinite loop?

Or it could be a glitch on the github site.

-brian

galran (approved):

PR #228, opened 2021-07-11
Upgrade to a new version of Makie, GLMakie and WGLMakie

galran:
This PR is the next step in supporting the new Makie/AbstractPlotting interface.
We are using the new Version of Makie: 0.14 and the new versions of GLMakie@0.4.2 and WGLMakie@0.4.2.

Upgrade: Makie@0.14, GLMakie@0.4, WGLMakie@0.4
Added some utility functions for the documentation - currently added them in NotebooksUtilities as they are very similar to the utility functions to support Pluto.
Removed support for older version of Makie due to the fact that we are limiting the version in the project.toml
Seems that the new Makie draw arrows in figures a bit differently than previous versions which cause some changes in some debug features in the Emitters module and a few function in the Vis module.
Changed the support for the 3D, 2Dx and 2Dy buttons in the Makie figures - need to be checked well.
Added first try at producing documentation figures that are interactable - currently just in the Emitters documentation page for 3D drawings.


src/NotebooksUtils/NotebooksUtils.jl
```diff
@@ -248,28 +251,57 @@ function SetBackend(defs::Defs, be::String)
...
     end
+end
 
+"""
+    function PrepareForDocs()
+
+Internal function!!! Override some code functions to be able to produce interactive figure for the documentation.
+"""
+function PrepareForDocs()
+    @eval begin
```
BrianGun, line 289: why does this have to be eval'ed?

galran: because i'm overriding the "natural behavior" of the Vis module and i want to do it only when i call this function (when we generate the documentation).
After this function execute, the Vis scene creation is going to use this function without the ability to return to the original behavior.
A better solution would be to change the Vis function to support better control over the way it works with the Makie  library but i don't want to touch that one now.

BrianGun: Is the presumption that this function will only be called when we make the documentation on github? Then it's safe because nothing other than documentation stuff is called in the process that makes the documentation, is that the idea?

@galran What happens when you make the documentation locally? Does the function redefinition persist afterward? That could be very confusing to newcomers, and to us a few months from now after we've forgotten about this redefinition. Is there a way to create a local context for the redefinition? Wrap it in a try block and somehow get hold of the original definition, stash it in a temp and then restore it after the documentation is built, or if the function crashes while building the documentation?

BrianGun: @galran 

there are 5 Makie.display(x::SomeType)  methods:
```
[9] display(re::RecordEvents) in Makie at /home/brian/.julia/packages/Makie/Riyar/src/display.jl:311
[10] display(scene::Scene) in Makie at /home/brian/.julia/packages/Makie/Riyar/src/display.jl:54
[11] display(fig::Figure) in Makie at /home/brian/.julia/packages/Makie/Riyar/src/display.jl:52
[12] display(fap::Makie.FigureAxisPlot) in Makie at /home/brian/.julia/packages/Makie/Riyar/src/display.jl:51
[13] display(x) in Base.Multimedia at multimedia.jl:324
```

when you define a new method 
```
Make.display(obj) = return obj
```
the type of obj is implicitly Any. So is this function meant to override method [13] in the list above? I would expect that if display is called with any of the more strictly typed arguments in [9]-[12] then those methods would be called instead of the redefined method, so effectively they are not redefined.

Is this what you intended? Just asking, because I don't fully understand what the code is doing.

BrianGun: Here's how I got the methods for Makie.display:

```
julia> methods(Makie.display)
# 13 methods for generic function "display":
[1] display(d::AbstractDisplay, mime::AbstractString, x) in Base.Multimedia at multimedia.jl:216
[2] display(mime::AbstractString, x) in Base.Multimedia at multimedia.jl:217
[3] display(d::TextDisplay, M::MIME{Symbol("text/plain")}, x) in Base.Multimedia at multimedia.jl:242
[4] display(d::TextDisplay, M::MIME, x) in Base.Multimedia at multimedia.jl:247
[5] display(d::TextDisplay, x) in Base.Multimedia at multimedia.jl:243
[6] display(m::MIME, x) in Base.Multimedia at multimedia.jl:338
[7] display(d::REPL.REPLDisplay, mime::MIME{Symbol("text/plain")}, x) in REPL at /home/brian/Applications/julia-1.6.1/share/julia/stdlib/v1.6/REPL/src/REPL.jl:211
[8] display(d::REPL.REPLDisplay, x) in REPL at /home/brian/Applications/julia-1.6.1/share/julia/stdlib/v1.6/REPL/src/REPL.jl:225
[9] display(re::RecordEvents) in Makie at /home/brian/.julia/packages/Makie/Riyar/src/display.jl:311
[10] display(scene::Scene) in Makie at /home/brian/.julia/packages/Makie/Riyar/src/display.jl:54
[11] display(fig::Figure) in Makie at /home/brian/.julia/packages/Makie/Riyar/src/display.jl:52
[12] display(fap::Makie.FigureAxisPlot) in Makie at /home/brian/.julia/packages/Makie/Riyar/src/display.jl:51
[13] display(x) in Base.Multimedia at multimedia.jl:324
```
or do this if you want to narrow it down to methods that have at least one argument of a particular type:
```
julia> methodswith(MIME,Makie.display)
[1] display(d::TextDisplay, M::MIME, x) in Base.Multimedia at multimedia.jl:247
[2] display(m::MIME, x) in Base.Multimedia at multimedia.jl:338
```

galran: Oh, I didn't know that they so many display types.
I assumed it's only for a Scene. I will fix the definition to override the scene one.

BrianGun (approved):

PR #244, opened 2021-07-19
Fix doc build timeout

### LINKED ISSUE #243: Doc build workflow timing out
e.g. https://github.com/microsoft/OpticSim.jl/actions/runs/1039605586

~This seems to have started since upgrading to Makie v0.15. Will try reverting the compat entry.~

Doesn't seem to be a Makie issue - the action before the upgrade also timed out after re-running...

Similar issue to #62

alfredclwong:
downgrade Makie v0.15 -> v0.14

fixes #243


BrianGun:
@alfredclwong any idea what is causing the documenter.jl build to take forever and eventually time out?

BrianGun:
@alfredclwong other people seem to be having similar problems: https://github.com/JuliaDocs/Documenter.jl/issues/1652. It's probably worth filing an issue on the documenter.jl github site so they realize this is widespread and prioritize a fix.

PR #249, opened 2021-07-26
Re-add GlassCat.EYE module

alfredclwong:
Had to dig this up from the old repo and move from AGF to OTHER. Should be useful for upcoming eye stuff. Also required for current eye model.


BrianGun:
@alfredclwong CARGILLE is not defined which is causing all the tests to fail.

alfredclwong:
> @alfredclwong CARGILLE is not defined which is causing all the tests to fail.

Missed this, thanks. Should be fixed now.

BrianGun (approved):

PR #258, opened 2021-08-05
Galran/adding projected paraxial lenses

galran:
added the HeadEye module inside the ParaxialAnalysis module to support the creation of Paraxial MLAs based on projection onto a Plane, Cylinder or Sphere.

This is a draft version - comments and exports will be added after initial integration to better support the tools and interfaces we will need.


src/Optical/ParaxialAnalysis/HeadEyeModel/Misc.jl
```diff
@@ -0,0 +1,76 @@
...
+"""
+function plane_from_points(points::Vector{SVector{3, Float64}}) 
+    if length(points) < 3 
+        return nothing      # At least three points required
+    end
+
+    sum = Base.sum(points)
+    centroid = Vec3(sum * (1.0 / (length(points))))
+
+    # Calc full 3x3 covariance matrix, excluding symmetries:
```
BrianGun, line 17: there is a function Statistics.cov which I think computes the same thing. Would be simpler to use it, unless you are doing something non-standard. We already use the Statistics package so this wouldn't add a dependency.

src/Optical/ParaxialAnalysis/HeadEyeModel/Misc.jl
```diff
@@ -0,0 +1,76 @@
...
+    end
+
+    # Pick path with best conditioning:
+    if      (det_max == det_x)
+        dir = Vec3(det_x,xz*yz - xy*zz,xy*yz - xz*yy)
+    elseif  (det_max == det_y)
+        dir = Vec3(xz*yz - xy*zz, det_y, xy*xz - yz*xx)
+    else
+        dir = Vec3(xy*yz - xz*yy,xy*xz - yz*xx,det_z)
+    end
```
BrianGun, line 51: Is this computation time sensitive? If not then it might be simpler to use the SVD on a matrix of the points and use the singular vectors as the coordinate frame of the best fitting plane. This would only be a few lines of code:

y= x .- mean(x,dims=2)
U,_,_ = svd(y)

then choose which of the U vectors you want to align with the x,y,z axes. Or just return the largest singular vector. 

Built in svd on 3x100 Matrix takes 10usec. If you use MMatrix instead it takes 6us. Or you can use my JacobiSVD (available on request) on MMatrix and it takes 3usec.

src/Geometry/Primitives/NonCSG/ConvexPolygon.jl
```diff
@@ -168,11 +167,11 @@ end
 
 Create a triangle mesh that can be rendered by iterating on the polygon's edges and for each edge use the centroid as the third vertex of the triangle.
 """
-function makemesh(poly::ConvexPolygon{N,T}, ::Int = 0) where {N,T<:Real}
+function makemesh(poly::ConvexPolygon{T,N}, ::Int = 0) where {N,T<:Real}
     c = centroid(poly)
 
     l2w = local2world(poly.local_frame)
-    len = length(poly.local_points)
+    len = size(poly.local_points)[2]
```
BrianGun, line 174: Is this right? local_points is a Vector{SVector{2,T}}. 
```
size(poly.local_points)[2]
```
gives an indexing error on my machine.

src/Optical/ParaxialAnalysis/HeadEyeModel/Arrangement.jl
```diff
@@ -0,0 +1,148 @@
...
+
+    function Shape(center::SVector{N, T}, points::Vector{SVector{N, T}}, coordinates::Tuple{Int,Int}) where {N, T<:Real}
+        return new{N, T}(center, points, coordinates)
+    end
+end
+
+center(h::Shape) = h._center
+points(h::Shape) = h._points
+coordinates(h::Shape) = h._coordinates
+
```
BrianGun, line 20: This is more idiomatic Julia
```
getshapes(::Type{Any};blah blah) = @error "Unknown Type ...."
getshapes(::Type{Rectangle}; blah blah) = return get_rectangles(resolution,size)
getshapes(::Type{Hexagon}; blah blah) = return get_hexagons(resolution,size)

File: src/Optical/ParaxialAnalysis/HeadEyeModel/Arrangement.jl [hunk 5]
```diff
@@ -0,0 +1,148 @@
...
+        points = [(SVector(p...) + center) for p in eachrow(basic_tile)]
+        center = center
+        push!(res, Shape(center, points, c))
+    end
+    return res
+end
+
+
+function project(shapes::Vector{Shape{2, T}}, csg::OpticSim.CSGTree{T})::Vector{Shape{3, T}} where {T<:Real}
+    ray_origin_distance = -10.0
```
BrianGun, line 76: should ray_origin_distance be an argument to this function? Or does it make sense for it to always be the same?

src/Geometry/Primitives/NonCSG/ConvexPolygon.jl
```diff
@@ -21,7 +21,7 @@ The local frame defines the plane (spans by the right and up vectors) with the p
 the local_polygon_points are given with respect to the local frame and are 2D points.
 NOTE: This class uses static vectors to hold the points which will lead to more efficient performance, but should not be used with polygons with more than 20-30 points.
 """
-struct ConvexPolygon{N,T<:Real}  <: PlanarShape{T} 
+struct ConvexPolygon{T<:Real, N}  <: PlanarShape{T}
```
BrianGun, line 24: was there some reason you changed the order of T and N? I did it this way so it would be similar to the StaticArrays type declarations. Does having N first cause problems with the code?

BrianGun (commented):
@galran could you look at the comments on the individual files? github is nagging me to leave this comment for you.

src/Optical/ParaxialAnalysis/HeadEyeModel/Misc.jl
galran: will do.

src/Optical/ParaxialAnalysis/HeadEyeModel/Misc.jl
galran: i'm not sure. i had some stability issues in the past with SVG in extreme cases - not sure if because of numerical errors or other reason. will check it out.

src/Geometry/Primitives/NonCSG/ConvexPolygon.jl
galran: will check.

src/Optical/ParaxialAnalysis/HeadEyeModel/Arrangement.jl
galran: agreed.

src/Optical/ParaxialAnalysis/HeadEyeModel/Arrangement.jl
galran: well, the assumption is that the original shapes are on the XY plane and the CSG surface to project is in the positive Z half space (above the XY plane), so it is safe to shoot from a negative Z position upwards.
IF the CSG surface is complex and moves into the negative Z it might present a problem.

src/Geometry/Primitives/NonCSG/ConvexPolygon.jl
galran: reversing this change after our conversation.

src/Optical/ParaxialAnalysis/HeadEyeModel/Misc.jl
BrianGun: @galran if you have stability issues then try my JacobiSVD code. It is supposed to compute small singular vectors to higher accuracy than the more common algorithms. And it is twice as fast as the Julia built in SVD.

src/Geometry/Primitives/NonCSG/ConvexPolygon.jl
galran: poly.local_points is an SMatrix - that is why i changed it from length to size

BrianGun (approved):

BrianGun (approved):

BrianGun (approved):

BrianGun (approved):

PR #261, opened 2021-08-09
Downgrade Documenter to "< 0.27"

alfredclwong:
something in 0.27 is causing our doc build to hang. not really worth chasing down (i've tried).


BrianGun:
@alfredclwong did this fix the problem? I noticed that all the documentation files are commented out. Does the documentation build when they are uncommented?

alfredclwong:
> @alfredclwong did this fix the problem? I noticed that all the documentation files are commented out. Does the documentation build when they are uncommented?

I've just been chucking a load of scenarios at the server to see what works. Seems like I've found the fix - will clean this up and request a review when it's confirmed.

BrianGun (approved):

PR #273, opened 2021-08-24
LensAssembly x PlanarShape

### LINKED ISSUE #272: ParaxialLenses will not be unpacked from nested LensAssemblies
https://github.com/microsoft/OpticSim.jl/blob/351427da364ae3a0f4d74a2a068a8abea9286270/src/Optical/LensAssembly.jl#L190-L210

alfredclwong:
Minor changes to LensAssembly to make it easier to work with, e.g. combining Rectangles, Hexagons and Ellipses into one PlanarShape Vector.

Fixes #272


src/Optical/LensAssembly.jl
```diff
@@ -50,37 +53,16 @@ macro lensassembly_constructor(N)
...
-            ellipses::Vector{Ellipse{T}}
-            hexagons::Vector{Hexagon{T}}
+            planarshapes::Vector{PlanarShape{T}}
             paraxials::Vector{ParaxialLens{T}}
             $(fields...)
         end
     end)
 end
 
-# This is the basic idea of what each closestintersection implementation does:
```
friggog, line 62: I left this commented code originally because it is quite hard to parse the logic just from reading through the macros below - for someone new to the codebase I think it is useful to keep it

src/Optical/LensAssembly.jl
```diff
@@ -170,55 +126,47 @@ macro lensassembly_elements(N)
...
+    if !isdefined(OpticSim, :($(Symbol("LensAssembly$N"))))
         @eval @lensassembly_constructor($N)
         @eval @lensassembly_intersection($N)
         @eval @lensassembly_elements($N)
     end
 end
 
-function LensAssembly(elements::Vararg{Union{Surface{T},CSGTree{T},LensAssembly{T}}}; axis::SVector{3,T} = SVector{3,T}(0.0, 0.0, 1.0)) where {T<:Real}
+# pregenerate a small number at compile time
+compile_lensassembly.(1:PREGENERATED_LENS_ASSEMBLY_SIZE)
```
friggog, line 138: Much neater!

src/Optical/LensAssembly.jl
```diff
@@ -170,55 +126,47 @@ macro lensassembly_elements(N)
...
+
+function LensAssembly(elements::Vararg{LensAssemblyElement{T}}; axis::SVector{3,T} = SVector{3,T}(0.0, 0.0, 1.0)) where {T<:Real}
     # make the actual object
-    actual_elements = []
-    rectangles = Vector{Rectangle{T}}(undef, 0)
```
friggog, line 186: I think I did this originally to avoid abstract types and consequent allocations - maybe this isn't relevant anymore?
The typed elements get explicitly parameterized in the LensAssembly type so are ok, but here I would think the array of `PlanarShape{T}` (which is abstract) results in tracing objects of unspecified type in `closest_intersection(lens_assembly, ray)`

test/testsets/Allocations.jl
```diff
@@ -5,6 +5,6 @@
 @testset "Allocations" begin
     # ensure that there are 0 allocations for all the benchmarks
     for b in Benchmarks.all_benchmarks()
-        @test Benchmarks.runbenchmark(b, samples = 1, evals = 1).allocs == 0
+        @test Benchmarks.runbenchmark(b, samples = 1, evals = 1).allocs == 0 || "$b"
```
friggog, line 8: Won't this allow allocations?

alfredclwong: I found this somewhere as a trick to display a custom error message on test failure (previously, an allocation test would fail without telling you which benchmark was causing allocations). I've still had runs where the allocation tests fail with this code, so I think it's ok.

src/Optical/LensAssembly.jl
alfredclwong: It seems to pass the tests without allocations, so maybe it's ok now? Otherwise, are you saying we'll have to revert back to separate rectangle/ellipse/hexagon vectors?

src/Optical/LensAssembly.jl
alfredclwong: Fair enough, I can add it back in. Although I've actually further cleaned up the code in this other branch https://github.com/microsoft/OpticSim.jl/commit/ad066db43a0b0e70e8e4965c96ffc9203d358c0b. It's not quite the same though - do you think this is explanatory enough?

src/Optical/LensAssembly.jl
alfredclwong: Reading [this](https://docs.julialang.org/en/v1/manual/performance-tips/#man-performance-abstract-container), it seems that this change may or may not cause allocations, but it's at least less memory efficient. Is this an issue? Or are we happy as long as there are no allocations?

friggog: Runtime allocations (at least used to) result in performance that was many orders of magnitude slower, so providing allocs stay at  you're probably ok even if the code seems less efficient. You can always look at the benchmark time as well as the allocs as a sanity check

src/Optical/LensAssembly.jl
```diff
@@ -170,55 +126,47 @@ macro lensassembly_elements(N)
...
+        elseif e isa LensAssembly{T}
             # unpack the lens assembly
-            rectangles = vcat(rectangles, e.rectangles)
-            ellipses = vcat(ellipses, e.ellipses)
-            hexagons = vcat(hexagons, e.hexagons)
-            for lae in OpticSim.typed_elements(e)
-                push!(actual_elements, lae)
-            end
+            planarshapes = vcat(planarshapes, e.planarshapes)
+            paraxials = vcat(paraxials, e.paraxials)
```
alfredclwong, line 155: fixes #272

src/Optical/LensAssembly.jl
friggog:
```
Allocations: Error During Test at /home/runner/work/OpticSim.jl/OpticSim.jl/test/testsets/Allocations.jl:8
  Expression evaluated to non-Boolean
  Expression: (Benchmarks.runbenchmark(b, samples = 1, evals = 1)).allocs == 0 || "$(b)"
       Value: "planar_shapes"
```
😢

friggog: > Do you remember if there are tests targeting allocs on rectangle/ellipse/hexagon-based LensAssemblies?

Really great catch on this - narrowly avoided a major performance regression due to my oversight when writing the alloc tests initially!

alfredclwong: :'(

At least we know the test works!

friggog: Yeah this sounds like a sensible idea - it seems redundant to have them all as separate implementations in ConvexPolygon can do the same thing. I guess Ellipse may have a purpose still, also would be interested in performance comparison between e.g. Rectangle and ConvecPolygon - though I expect there will be little difference

BrianGun:
@alfredclwong the allocations test is still failing so I assume this is not ready to be approved?

PR #283, opened 2021-09-08
Object wrapper type

alfredclwong:
Introduce the ability to tag any variable with an `id::UUID` field, allowing for more object-oriented behaviours within OpticSim.

https://github.com/microsoft/OpticSim.jl/blob/87a61c701f005a203a26f0805c128626c032f062/src/Object.jl#L9-L14

This implementation preserves backwards compatibility: due to multiple dispatch, existing functions for a given type `T` will work as before, but we can now write new functions for `Object{T}` to specify enhanced behaviours. For example, the following new `draw!` code in Visualization.jl will extract color properties from the object before unpacking and delegating to another existing `draw!` function.

https://github.com/microsoft/OpticSim.jl/blob/235a61e1404d30d3833951d8175dcb2f42986c72/src/Vis/Visualization.jl#L271-L274

Example usage of this feature can be seen in `draw_lensconstruction`.

https://github.com/microsoft/OpticSim.jl/blob/87a61c701f005a203a26f0805c128626c032f062/src/Examples/docs_examples.jl#L133-L139

Object types and non-Object types can be used in tandem - everything has been written in order to cause minimal disruption to existing code. However, OpticSim has a very broad codebase, and there are many more functions that will need extending to support full usage of Object types in all possible scenarios. This PR includes a small subset of extensions for demonstrative purposes, but we can add to this over time (e.g. for multidetectors).


src/Optical/LensAssembly.jl
```diff
@@ -30,127 +35,79 @@ Where `Ti <: Union{Surface{T},CSGTree{T}}`.
...
         struct $(name){T,$(typesig...)} <: LensAssembly{T}
             axis::SVector{3,T}
-            rectangles::Vector{Rectangle{T}}
-            ellipses::Vector{Ellipse{T}}
-            hexagons::Vector{Hexagon{T}}
-            paraxials::Vector{ParaxialLens{T}}
+            rectangles::Vector{unionobject(Rectangle{T})}
+            ellipses::Vector{unionobject(Ellipse{T})}
+            hexagons::Vector{unionobject(Hexagon{T})}
+            paraxials::Vector{unionobject(ParaxialLens{T})}
```
alfredclwong, line 63: This could be causing allocations - will look into it when I can

src/Object.jl
```diff
@@ -0,0 +1,37 @@
...
+struct Object{T}
+    object::T
+    id::UUID
+
+    Object(object::T, id::UUID = uuid()) where {T} = new{T}(object, id)
+end
+
+Base.show(io::IO, object::Object) = print(io, "[$(object.id)]")
+
+unionobject(T::Type) = Union{T, Object{<:T}}
```
friggog, line 22: I don't think the compiler will be able to reason about types that result from a function statically - I think using macros for this purpose would work better

src/Optical/Paraxial.jl
```diff
@@ -17,11 +17,11 @@ ParaxialLensConvexPoly(focaldistance, local_frame, local_polygon_points, local_c
 ```
 """
 struct ParaxialLens{T} <: Surface{T}
-    shape::Union{Rectangle{T},Ellipse{T},Hexagon{T},ConvexPolygon{N, T} where {N}} 
+    shape::PlanarShape{T}
```
friggog, line 20: Using an abstract type here might mean that the type of shape can't be resolved when tracing - you could parameterise ParaxialLens with the shape type or dimensionality as well as type maybe?

friggog:
It might be worth adding benchmarks for surface intersection of planarshape
Currently we just have tracing of a system with these in:
`planar_shapes() = trace, (TestData.planarshapes(), optray())`
It could be useful to decouple this from allocs just in intersecting the shape directly similar to this:
`rectangle() = surfaceintersection, (Rectangle(1.0, 1.0, SVector(0.0, 0.0, 1.0), SVector(0.0, 0.0, 0.0)), rayz())`

src/Object.jl
friggog: Although seems ok given that only the planar shapes test is causing allocs

alfredclwong:
Thanks for the review @friggog, I agree with the comments and will take them into account. Closing this PR for now as I'm focusing on another task. Will get back to this later in the week with another rewrite most likely.

PR #289, opened 2021-09-15
Fix runforpipeline

### LINKED ISSUE #288: Fix runforpipeline
runforpipeline used [].join which apparently doesn't work anymore. Changed to join(io,[])

BrianGun:
Fixes #288

old code used [].join(", ") which doesn't appear to work anymore. Changed to join([],", "). Manually tested the function to verify it worked.


alfredclwong (approved):
I think I wrote this recently. Not sure how this slipped through, thanks for correcting it!

PR #325, opened 2021-11-24
added the InfiniteStopConvexPoly struct to support convex poly aperture

galran:
added the RectJitterGrid to the Emitters.Origins to support better (less clamping) way of sampling
added a convex polygon Stop shape to support the hexagon shaped lenses


friggog:
Shouldn’t this check for intersection with the poly as well and return empty if intersection and the plane intersection interval otherwise?

friggog:
Emitter change also seems unrelated to this PR

galran:
> Shouldn’t this check for intersection with the poly as well and return empty if intersection and the plane intersection interval otherwise?

agreed. the code seems weird - will check and maybe split this PR into two

BrianGun (approved):

friggog:
Replaced by #328

PR #328, opened 2021-11-29
added the InfiniteStopConvexPoly shape to allow aperture for hex lenses.

galran:
also fixed a small bug in convex polygon assert statement.

### DIFF
--- a/src/Geometry/Primitives/NonCSG/ConvexPolygon.jl
+++ b/src/Geometry/Primitives/NonCSG/ConvexPolygon.jl
@@ -38,7 +38,7 @@ struct ConvexPolygon{N,T<:Real}  <: PlanarShape{T}
         ) where {T<:Real}
 
         # need at least 3 points to define apolygon
-        @assert length(local_polygon_points) > 3         
+        @assert length(local_polygon_points) >= 3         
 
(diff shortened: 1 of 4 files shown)


friggog (approved):
Maybe should an allocation test for the new stop just to make sure?

galran:
> 

do i need to add an allocation test explicitly? i thought that they run for all the benchmarks.

brian-guenter:
> > 
> 
> do i need to add an allocation test explicitly? i thought that they run for all the benchmarks.

@galran if you add a new type you need to add an allocation test. Check out test/Benchmarks/Benchmarks.jl. You add your test code to list of tests to perform and then it will be run. This is something we shouldn't skip - if we let allocations creep in it will be difficult to figure out where performance hits are coming from later on.

galran:
added allocation tests. hope i did it correctly.

brian-guenter (approved):

brian-guenter:
> added allocation tests. hope i did it correctly.

I looked them over. They look correct. And they all passed!

brian-guenter (approved):

brian-guenter (approved):

BrianGun (approved):

PR #415, opened 2022-10-18
Update GridSag.jl

gerritlaube:
The function "GridSagSurface(... AbstractArray{T,2}...)" tries to assert that the third dimension of the array has length 4, althoug the input only has two dimensions. The third dimension should be filled with all zeroes in line 46. Probably copy-paste issue from the 3D-array functions.


friggog (approved):

friggog:
Thanks for the fix @gerritlaube!

